PR #7, opened 2020-09-24
Functors task 1 solution.

kolya-kobets:


functors task (1)/student/include/impl.h
```diff
@@ -1,13 +1,27 @@
...
+    BeerBrand operator()()
+    {
+        if(mCurrentBrand == OrderedBrands.end())
+        {
+            mCurrentBrand = OrderedBrands.begin();
+        }
+        return *(mCurrentBrand++);
+    }
+private:
+    static const std::vector<BeerBrand> OrderedBrands;
```
smay1613, line 23: You're right, but here we have guarantee about the enum values order (it won't be changed).
This is real-world solution. Task solution is simpler.
But it is still accepted!

functors task (1)/student/include/impl.h
```diff
@@ -40,9 +56,16 @@ struct MixingPolicy
      * Whiskey + SevenUp = SevenPlusSeven;
      * Others + Others = Oops;
      */
-    static Cocktail mix(/**???*/)
+    static Cocktail mix(AlcoholDrink alcohol, NonAlcoholDrink non_alcohol)
     {
+        static const std::map<std::pair<AlcoholDrink, NonAlcoholDrink>, Cocktail> recipes = {
```
smay1613, line 61: Beautiful solution :)
Mapping is the best choice here.
Instead of a long switch, you should prefer value mapper.

PR #8, opened 2020-09-24
functors task (1) - Hnoinskyi Artur

ArchiGn98:
Functors task 1 is completed with 13/15 tests passed

### DIFF
--- a/functors task (1)/student/include/impl.h
+++ b/functors task (1)/student/include/impl.h
@@ -8,37 +8,15 @@
  */
 struct BeerOrganizer
 {
-  size_t current_index {0};
-  std::array<BeerBrand, 6> beer_arr {
-        BeerBrand::HoeGaarden,
-        BeerBrand::Corona,
-        BeerBrand::Carlsberg,
-        BeerBrand::Bud,
-        BeerBrand::ZlataPraha,
-        BeerBrand::Leffe
-  };
-
+  size_t current_index {static_cast<size_t>(BeerBrand::HoeGaarden)};
 
   BeerBrand operator() (){
-    if(current_index == beer_arr.size())
+    if(current_index == static_cast<size_t>(BeerBrand::Max))
       {
-        current_index = 0;
+        current_index = static_cast<size_t>(BeerBrand::HoeGaarden);
       }
-    return beer_arr[current_index++];
+    return static_cast<BeerBrand>(current_index++);
   }
-
-};
-
-enum class lol
-{
-    None = 0,
-    HoeGaarden,
-    Corona,
-    Carlsberg,
-    Bud,
-    ZlataPraha,
-    Leffe,
-    Max = 7
 };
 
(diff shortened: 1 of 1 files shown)


functors task (1)/student/include/impl.h
```diff
@@ -8,6 +8,37 @@
...
+    if(current_index == beer_arr.size())
+      {
+        current_index = 0;
+      }
+    return beer_arr[current_index++];
+  }
+
+};
+
+enum class lol
```
smay1613, line 32: please don't commit unrelated (debug also) changes!

functors task (1)/student/include/impl.h
```diff
@@ -8,6 +8,37 @@
  */
 struct BeerOrganizer
 {
+  size_t current_index {0};
+  std::array<BeerBrand, 6> beer_arr {
```
smay1613, line 12: everything is simpler, because we have a guarantee about beer order in enum

functors task (1)/student/include/impl.h
```diff
@@ -16,16 +47,31 @@ struct BeerOrganizer
  *
  * @note Only Corona and HoeGaarden are expensive
  */
-bool isExpensiveBeer(/**???*/)
+bool isExpensiveBeer(BeerBrand brand)
 {
+  switch(brand){
```
smay1613, line 52: please never use switches in such way :)

if your return value is duplicated, you can use fallthrough.
But, for these two types it is enough to use comparing like:

return brand == BeerBrand::Corona || brand == BeerBrand::HoeGaarden

functors task (1)/student/include/impl.h
```diff
@@ -40,9 +86,24 @@ struct MixingPolicy
      * Whiskey + SevenUp = SevenPlusSeven;
      * Others + Others = Oops;
      */
-    static Cocktail mix(/**???*/)
+    static Cocktail mix(AlcoholDrink alco, NonAlcoholDrink nonAlco)
     {
+      if(alco == AlcoholDrink::Gin && nonAlco == NonAlcoholDrink::LimeJuice)
```
smay1613, line 91: You can simplify these checks with switch

smay1613 (changes requested):

functors task (1)/student/include/impl.h
```diff
@@ -88,21 +53,19 @@ struct MixingPolicy
...
-        {
-          return Cocktail::SevenPlusSeven;
-        }
-      else{
-          return Cocktail::Oops;
-        }
+      switch(alco){
+        case AlcoholDrink::Gin:
+          {
+            if(nonAlco == NonAlcoholDrink::LimeJuice) return Cocktail::Gimlet;
```
smay1613, line 59: please don't write two expressions on one line
also, https://wiki.c2.com/?AlwaysUseBracesOnIfThen

PR #9, opened 2020-09-24
Implement functors task #1 - ENikonov

re-nikonov:


functors task (1)/student/include/impl.h
```diff
@@ -40,9 +45,16 @@ struct MixingPolicy
      * Whiskey + SevenUp = SevenPlusSeven;
      * Others + Others = Oops;
      */
-    static Cocktail mix(/**???*/)
+    static Cocktail mix(AlcoholDrink alcoholDrink, NonAlcoholDrink nonAlcoholDrink)
     {
+        if (alcoholDrink == AlcoholDrink::Gin && nonAlcoholDrink == NonAlcoholDrink::LimeJuice)
```
smay1613, line 50: Please fix formatting!
Always use braces after loop/if:
https://stackoverflow.com/questions/2125066/is-it-a-bad-practice-to-use-an-if-statement-without-curly-braces

functors task (1)/student/include/impl.h
```diff
@@ -40,9 +45,16 @@ struct MixingPolicy
...
+            return Cocktail::Gimlet;
+        if (alcoholDrink == AlcoholDrink::Gin && nonAlcoholDrink == NonAlcoholDrink::GrapefruitJuice)
```
smay1613, line 52: you can simplify these checks with switch

functors task (1)/student/src/impl.cpp
```diff
@@ -1,2 +1,16 @@
 #include "impl.h"
 
+static constexpr BeerBrand kBeerBrandsOrdered[] = {
```
smay1613, line 3: **never use c array in C++**
use std::array instead!

functors task (1)/student/src/impl.cpp
```diff
@@ -1,2 +1,16 @@
...
+    BeerBrand::HoeGaarden,
+    BeerBrand::Corona,
+    BeerBrand::Carlsberg,
+    BeerBrand::Bud,
+    BeerBrand::ZlataPraha,
+    BeerBrand::Leffe,
+};
+static constexpr auto kBeerBrandsOrderedSize = sizeof(kBeerBrandsOrdered) / sizeof(kBeerBrandsOrdered[0]);
```
smay1613, line 11: std::array has .size() method

functors task (1)/student/src/impl.cpp
```diff
@@ -1,2 +1,16 @@
...
+
+BeerBrand BeerOrganizer::operator()()
+{
+    return kBeerBrandsOrdered[i++ % kBeerBrandsOrderedSize];
```
smay1613, line 15: everything is much simpler:
we have a guarantee about enum values order here

you shouldn't use array

smay1613 (changes requested):

functors task (1)/student/src/impl.cpp
re-nikonov: That's true. Array removed.

functors task (1)/student/src/impl.cpp
re-nikonov: Never say never :)
There are some pros and cons of using std::array vs C-array.
1. std::array must specify size, it is not flexible, but yes, it is much better than do "sizeof() / sizeof()", my bad... until C++17.
2. Both can be passed to algorithms as they have std::begin() and std::end().
3. Both can be passed to other [template] functions.
4. constexpr int array[] = { 1, 2, ..., N }; looks much simpler and doesn't suffer from (1).
In general, I'd say that from C++17 (where we've got std::size() for both of them) there are not much advantages in std::array to declare const[expr] arrays... They are quite similar.

Anyway, this one was removed.

functors task (1)/student/include/impl.h
re-nikonov: Am I understand you right? Do you want to say that two nested switches (or switch with if's) are more readable than cascaded if's. Don't think so... both are ugly. In case of much bigger number of cocktails there should've been some kind of predefined array or map and just call find() here, but for 3 cocktails, I'm not sure this is really necessary.

functors task (1)/student/include/impl.h
re-nikonov: In general it is true.

But what about different coding styles which allow you (and even encourage) to do so?
https://google.github.io/styleguide/cppguide.html#Conditionals
https://webkit.org/code-style-guidelines/#braces-one-line
https://llvm.org/docs/CodingStandards.html#don-t-use-braces-on-simple-single-statement-bodies-of-if-else-loop-statements
Google, WebKit, LLVM... they are not so stupid to miss that discussion on stack overflow. What do you think?

Yes, I'll fix it. But in general, if we don't have dedicated coding style for this training, I'd say it is not the best idea to say "always" here.

functors task (1)/student/include/impl.h
```diff
@@ -8,6 +8,9 @@
  */
 struct BeerOrganizer
 {
+    BeerBrand operator()();
+
+    size_t i = 0;
```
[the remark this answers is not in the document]
re-nikonov: :) my bad, wanted to finish faster... bad practice in professional life.

functors task (1)/student/include/impl.h
smay1613: Yes, you're right about the map usage in this task (it's the best solution).
We can easily transform such kind of switch to a map (it will be discussed in "Containers" training).
Imho, switching logic is closer to mappng pattern, than such if, because you can see all pairs of values.

functors task (1)/student/src/impl.cpp
smay1613: You must specify size for any kind of array, because VLA is not C++ way.
But consider advantages of std::array (it is not worse than a c array):
* You can always access size
* Value semantics
* No casts to a pointer
* Range checking

functors task (1)/student/include/impl.h
```diff
@@ -1,13 +1,23 @@
...
 /**
  * @todo Implement functor-generator that will return next beer brand (cyclic)
  * @example HoeGaarden -> Corona -> Carlsberg -> Bud -> ZlataPraha -> Leffe -> HoeGaarden...
  */
 struct BeerOrganizer
 {
+    BeerBrand operator()()
+    {
+        if (nextBeerBrand == static_cast<int>(BeerBrand::Max)) {
+            nextBeerBrand = 1;
```
smay1613, line 15: magic number :)

functors task (1)/student/include/impl.h
```diff
@@ -40,9 +52,18 @@ struct MixingPolicy
      * Whiskey + SevenUp = SevenPlusSeven;
      * Others + Others = Oops;
      */
-    static Cocktail mix(/**???*/)
+    static Cocktail mix(AlcoholDrink alcoholDrink, NonAlcoholDrink nonAlcoholDrink)
     {
+        static const std::map<std::pair<AlcoholDrink, NonAlcoholDrink>, Cocktail> kCocktailsMap = {
```
smay1613, line 57: Best solution :)

functors task (1)/student/include/impl.h
re-nikonov: Fixed.

PR #10, opened 2020-09-24
Functors task (1) - Poplavskyi

NetLagina:


functors task (1)/student/include/impl.h
```diff
@@ -8,6 +8,11 @@
  */
 struct BeerOrganizer
 {
+    private:
+        uint currentBeerIndex = 1;
```
smay1613, line 12: _prefer size_t_

functors task (1)/student/include/impl.h
```diff
@@ -40,9 +47,20 @@ struct MixingPolicy
      * Whiskey + SevenUp = SevenPlusSeven;
      * Others + Others = Oops;
      */
-    static Cocktail mix(/**???*/)
+    static Cocktail mix(AlcoholDrink alcohol, NonAlcoholDrink nonAlcohol)
     {
+        Cocktail cocktail;
+        if (alcohol == AlcoholDrink::Gin && nonAlcohol == NonAlcoholDrink::LimeJuice) {
```
smay1613, line 53: you can simplify these checks with switch

functors task (1)/student/src/impl.cpp
```diff
@@ -1,2 +1,11 @@
 #include "impl.h"
 
+BeerBrand BeerOrganizer::operator()()
+{
+    if (currentBeerIndex == static_cast<int>(BeerBrand::Max))
+    {
+        currentBeerIndex = 1;
```
smay1613, line 7: _nice, but still magic counters in code_

smay1613 (changes requested):

functors task (1)/student/include/impl.h
NetLagina: changed

functors task (1)/student/src/impl.cpp
NetLagina: ok, changed to enum cast

smay1613 (commented):
Nice!

PR #12, opened 2020-09-24
functors task (1) - Sakun

AndreySakun:


functors task (1)/student/src/impl.cpp
```diff
@@ -1,2 +1,22 @@
 #include "impl.h"
 
+constexpr int count = 6;
+BeerBrand states[count] = { BeerBrand::HoeGaarden, BeerBrand::Corona, BeerBrand::Carlsberg, BeerBrand::Bud, BeerBrand::ZlataPraha, BeerBrand::Leffe };
```
smay1613, line 4: **never use c array in C++ code!**
use std::array instead!

functors task (1)/student/src/impl.cpp
```diff
@@ -1,2 +1,22 @@
 #include "impl.h"
 
+constexpr int count = 6;
```
smay1613, line 3: std::array has .size() method

functors task (1)/student/src/impl.cpp
```diff
@@ -1,2 +1,22 @@
...
+BeerBrand BeerOrganizer::operator()()
+{
+    BeerBrand ret = states[current%count];
+    current++;
+    return ret;
+}
+
+Cocktail MixingPolicy::mix(AlcoholDrink alcoDrink, NonAlcoholDrink nonAlcoDrink)
+{
+    if (alcoDrink == AlcoholDrink::Gin && nonAlcoDrink == NonAlcoholDrink::LimeJuice)
```
smay1613, line 15: use braces after if/loop:
https://wiki.c2.com/?AlwaysUseBracesOnIfThen

functors task (1)/student/src/impl.cpp
```diff
@@ -1,2 +1,22 @@
...
+        return Cocktail::Gimlet;
+    if (alcoDrink == AlcoholDrink::Gin && nonAlcoDrink == NonAlcoholDrink::GrapefruitJuice)
```
smay1613, line 17: you can simplify these checks with switch

smay1613 (changes requested):

functors task (1)/student/src/impl.cpp
AndreySakun: this method do not solve anything, because we still need to declare array with a constant:
std::array<BeerBrand, count>

std::make_array is still in proposal.
Honestly I wanted to use c-array with std::size, but this is available since c++17
Maybe there is better to use std::vector with initialization list

functors task (1)/student/src/impl.cpp
AndreySakun: completely agree

functors task (1)/student/src/impl.cpp
smay1613: Yes, but here it is still a magic number.
You can get size from casting enum values.

PR #17, opened 2020-09-24
Implementation of a Task

JohnDou1991:


smay1613:
Please fix pull request title

functors task (1)/student/src/impl.cpp
```diff
@@ -1,2 +1,33 @@
...
+    }
+
+    return static_cast<BeerBrand>( mLastBeerBrand++ );
+}
+
+bool isExpensiveBeer( const BeerBrand brand )
+{
+    std::array<BeerBrand,2> expensiveOnes = { BeerBrand::Corona, BeerBrand::HoeGaarden };
+    
+    for ( const BeerBrand expensive : expensiveOnes )
```
smay1613, line 17: in your case you can use std::initializer list as iteration range, but I think we don't need this loop

functors task (1)/student/src/impl.cpp
```diff
@@ -1,2 +1,33 @@
...
+    {
+        mLastBeerBrand = static_cast<size_t>( BeerBrand::None ) + 1;
+    }
+
+    return static_cast<BeerBrand>( mLastBeerBrand++ );
+}
+
+bool isExpensiveBeer( const BeerBrand brand )
+{
+    std::array<BeerBrand,2> expensiveOnes = { BeerBrand::Corona, BeerBrand::HoeGaarden };
```
smay1613, line 15: Nice idea, but for now it is better to keep it simple:
return brand == BeerBrand::Corona || brand == BeerBrand::HoeGaarden

functors task (1)/student/src/impl.cpp
```diff
@@ -1,2 +1,33 @@
...
+        if ( brand == expensive )
+            return true;
+    }
+
+    return false;
+}
+
+Cocktail MixingPolicy::mix( const AlcoholDrink& hard, const NonAlcoholDrink& soft )
+{
+    if ( hard == AlcoholDrink::Gin     && soft == NonAlcoholDrink::GrapefruitJuice ) return Cocktail::Greyhount;
```
smay1613, line 28: _for if/loop statement please always use braces:
https://wiki.c2.com/?AlwaysUseBracesOnIfThen_

smay1613, line 28: _please avoid two statements at one line_

smay1613 (changes requested):

PR #18, opened 2020-09-24
Functor task 1 - Diachenko

Rald1:


functors task (1)/student/include/impl.h
```diff
@@ -8,6 +8,10 @@
  */
 struct BeerOrganizer
 {
+    BeerOrganizer();
+    BeerBrand operator()();
+private:
+    int currentBeerBand;
```
smay1613, line 14: _prefer size_t here_

functors task (1)/student/src/impl.cpp
```diff
@@ -1,2 +1,22 @@
 #include "impl.h"
 
+BeerOrganizer::BeerOrganizer()
+   : currentBeerBand((int)BeerBrand::None)
```
smay1613, line 4: you can avoid this constructor and initialize in-place
**never use c-cast in c++ code**

functors task (1)/student/src/impl.cpp
```diff
@@ -1,2 +1,22 @@
...
+{}
+
+BeerBrand BeerOrganizer::operator()()
+{
+    if (++currentBeerBand == (int)BeerBrand::Max) {
```
smay1613, line 9: **never use c-cast in c++ code**

functors task (1)/student/src/impl.cpp
```diff
@@ -1,2 +1,22 @@
...
+        currentBeerBand = (int)BeerBrand::None;
+        currentBeerBand++;
```
smay1613, line 11: please re-consider this implementation, because it looks little strange

functors task (1)/student/src/impl.cpp
```diff
@@ -1,2 +1,22 @@
...
+    }
+
+    return (BeerBrand)currentBeerBand;
+}
+
+bool isExpensiveBeer(const BeerBrand beerBrand)
+{
+    if (beerBrand == BeerBrand::Corona || beerBrand == BeerBrand::HoeGaarden)
```
smay1613, line 19: why not return beerBrand == BeerBrand::Corona || beerBrand == BeerBrand::HoeGaarden?
you should avoid explicit true/false

smay1613 (changes requested):

functors task (1)/student/src/impl.cpp
```diff
@@ -1,22 +1,18 @@
...
 BeerBrand BeerOrganizer::operator()()
 {
-    if (++currentBeerBand == (int)BeerBrand::Max) {
-        currentBeerBand = (int)BeerBrand::None;
-        currentBeerBand++;
+    currentBeerBrand++;
+
+    if (currentBeerBrand == static_cast<size_t>(BeerBrand::Max)) {
+        currentBeerBrand = static_cast<size_t>(BeerBrand::None);
+        currentBeerBrand++;
```
smay1613, line 9: it looks beautiful now 👍

PR #21, opened 2020-09-24
functor task 1, VNevzorov

sam-mcdonalds:


functors task (1)/student/include/impl.h
```diff
@@ -1,48 +1,81 @@
...
- * @example HoeGaarden -> Corona -> Carlsberg -> Bud -> ZlataPraha -> Leffe -> HoeGaarden...
+ * @example HoeGaarden -> Corona -> Carlsberg -> Bud -> ZlataPraha -> Leffe ->
+ * HoeGaarden...
  */
-struct BeerOrganizer
-{
+struct BeerOrganizer {
+  // we can always be sure that generated beerBrand will be strictly between
+  // None and Max exclusive
+  BeerOrganizer() : m_beer{BeerBrand::None} {}
```
smay1613, line 17: Prefer in-place initialization and follow rule of zero

functors task (1)/student/include/impl.h
```diff
@@ -1,48 +1,81 @@
...
- * @todo Implement unary predicate function that will determine if beer is expensive
+ * @todo Implement unary predicate function that will determine if beer is
+ * expensive
  * @return true in case if beer is expensive, false if not
  *
  * @note Only Corona and HoeGaarden are expensive
  */
-bool isExpensiveBeer(/**???*/)
-{
+bool isExpensiveBeer(BeerBrand beer) {
```
smay1613, line 38: _don't forget about const :)_

functors task (1)/student/include/impl.h
```diff
@@ -1,48 +1,81 @@
...
+   * Gin + Lime juice = Gimlet;
+   * Gin + Grapefruit juice = Greyhount;
+   * Whiskey + SevenUp = SevenPlusSeven;
+   * Others + Others = Oops;
+   */
+
+  static std::map<std::pair<AlcoholDrink, NonAlcoholDrink>, Cocktail>
+      cocktailsRecipes;
+
+  static Cocktail mix(AlcoholDrink ad, NonAlcoholDrink nad) {
```
smay1613, line 65: _use more descriptive names, no abbreviations_

functors task (1)/student/include/impl.h
```diff
@@ -1,48 +1,81 @@
...
+    if (cocktailsRecipes.count({ad, nad}) > 0) {
```
smay1613, line 66: Map is a good alternative for a switch here! 
Nice :)

functors task (1)/student/include/impl.h
sam-mcdonalds: Does it is really needed when we passed arguments by value?
I mean, the caller is anyway guaranteed that the original value won't change.

smay1613: Just a style: always mark function params as const, because you can modify it accidentally

PR #25, opened 2020-09-24
Implement functors task #2 - ENikonov

re-nikonov:


functors task (2)/student/include/impl.h
```diff
@@ -62,10 +59,17 @@ class DataBrowser
...
+        const auto& it = m_dataReaders.find(userId);
+        if (it == m_dataReaders.cend())
+        {
+            return false;
+        }
+        if (it->second == nullptr)
+        {
+            return false;
+        }
+        return std::forward<T>(f)(it->second);
```
smay1613, line 72: _👍 for forward _

functors task (2)/student/src/impl.cpp
```diff
@@ -5,50 +5,41 @@ void DataBrowser::userLeave(const std::string &userId)
...
-    }
-    if (it->second == nullptr)
-    {
-        return false;
-    }
-    return it->second->getDataType1(returnValues, 0);
+    auto getDataTypeAdapter = [this, &returnValues](const std::unique_ptr<IDataSelector>& selector) {
+        using namespace std::placeholders;
+        return invokeDataRequest(
+            std::bind(&IDataSelector::getDataType1, _1, _2, 0), selector, returnValues);
```
smay1613, line 19: _acceptable, but prefer lambda_

functors task (2)/student/src/impl.cpp
```diff
@@ -5,50 +5,41 @@ void DataBrowser::userLeave(const std::string &userId)
...
-        return false;
-    }
-    if (it->second == nullptr)
-    {
-        return false;
-    }
-    return it->second->getDataType2(returnValues);
+    auto getDataTypeAdapter = [this, &returnValues](const std::unique_ptr<IDataSelector>& selector) {
+        return invokeDataRequest(
+            std::mem_fn(&IDataSelector::getDataType2), selector, returnValues);
```
smay1613, line 28: _lambda is more explicit_

re-nikonov: Implemented three different variants to see the difference.

smay1613: Nice!

re-nikonov: The idea about lambdas is clear. Used std::bind() to see visual difference. Anything else to fix?

smay1613: it's ok now

PR #27, opened 2020-09-24
functors task (2) - Sakun

AndreySakun:


functors task (2)/student/src/impl.cpp
```diff
@@ -7,48 +7,44 @@ void DataBrowser::userLeave(const std::string &userId)
...
+    auto func = [this, & returnValues](const std::unique_ptr<IDataSelector>& selector)
     {
-        return false;
-    }
-    if (it->second == nullptr)
-    {
-        return false;
-    }
-    return it->second->getDataType1(returnValues, 0);
+        auto f = [](const IDataSelector* selector, std::vector<size_t> &returnValues)
```
smay1613, line 12: _Such names are acceptable for templated code, but in future consider more descriptive names_

functors task (2)/student/src/impl.cpp
```diff
@@ -7,48 +7,44 @@ void DataBrowser::userLeave(const std::string &userId)
...
+    auto func = [this, & returnValues](const std::unique_ptr<IDataSelector>& selector)
     {
-        return false;
-    }
-    return it->second->getDataType2(returnValues);
+        auto f = [](const IDataSelector* selector, std::vector<size_t> &returnValues)
+        {
+            return selector->getDataType2(returnValues);
+        };
+        return invokeDataRequest(f, selector, returnValues);
```
smay1613, line 29: you can avoid wrapping here and call original function

PR #29, opened 2020-09-25
Functors: Task_1 Sarandi Ihor

IgorSarandi:
Signed-off-by: Sarandi Ihor <isarandi@luxoft.com>

### DIFF
--- a/functors task (1)/student/include/impl.h
+++ b/functors task (1)/student/include/impl.h
@@ -1,6 +1,7 @@
 #pragma once
 #include "bar_serving.h"
 #include <functional>
+#include <map>
 
 /**
  * @todo Implement functor-generator that will return next beer brand (cyclic)
@@ -8,6 +9,10 @@
  */
 struct BeerOrganizer
 {
+    BeerBrand operator()();
+
+    private:
+    size_t counter = static_cast<size_t>(BeerBrand::None);
 };
 
 /**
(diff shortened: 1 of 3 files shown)


functors task (1)/student/include/impl.h
```diff
@@ -1,13 +1,18 @@
...
 /**
  * @todo Implement functor-generator that will return next beer brand (cyclic)
  * @example HoeGaarden -> Corona -> Carlsberg -> Bud -> ZlataPraha -> Leffe -> HoeGaarden...
  */
 struct BeerOrganizer
 {
+    BeerBrand operator()() const;
+
+    private:
+    mutable int8_t counter = static_cast<int8_t>(BeerBrand::None);
```
smay1613, line 15: _prefer size_t_
why mutable?

functors task (1)/student/src/impl.cpp
```diff
@@ -1,2 +1,17 @@
...
+BeerBrand BeerOrganizer::operator()() const
+{
+    if((counter + 1) >= static_cast<int8_t>(BeerBrand::Max))
+    {
+        counter = static_cast<int8_t>(BeerBrand::None);
+    }
+    return static_cast<BeerBrand>(++counter);
+}
+
+std::map <std::pair<AlcoholDrink, NonAlcoholDrink>, Cocktail> MixingPolicy::bar =
```
smay1613, line 12: Mapping is the best solution here, so you're right.
It is more flexible switch logic.

smay1613 (changes requested):

functors task (1)/student/include/impl.h
IgorSarandi: mutable because I declare BeerBrand operator()() as const method

smay1613: I think mutable is not a good solution here.
In fact, you modify the internal state, so it shouldn't be const.

IgorSarandi: Done. Its make sense, thank.

PR #35, opened 2020-09-25
Functors task 2 - OSennikov

JohnDou1991:


functors task (2)/student/include/impl.h
```diff
@@ -21,7 +21,7 @@ class DataBrowser
      */
     void userEnter(const std::string& userId)
     {  // awful, but in header for testing
-        m_dataReaders.emplace(userId, getSelector());
+        m_dataReaders.emplace( userId, std::unique_ptr<IDataSelector>( getSelector() ) );
```
smay1613, line 24: cheating, please don't modify this code

or you mean something is wrong in task interface?

functors task (2)/student/src/impl.cpp
```diff
@@ -7,48 +7,47 @@ void DataBrowser::userLeave(const std::string &userId)
...
-    }
-    if (it->second == nullptr)
-    {
-        return false;
-    }
-    return it->second->getDataType1(returnValues, 0);
+        return selector->getDataType1( returnValues, 0 );
+    };
+
+    return safeCall( userId, [this, &returnValues, function]( const std::unique_ptr<IDataSelector>& selector ) -> bool
```
smay1613, line 15: Hard to read. Consider naming this lambda.
```suggestion
    auto getter = [this, &returnValues, function]( const std::unique_ptr<IDataSelector>& selector ) ...
    return safeCall( userId, getter)
```

functors task (2)/student/src/impl.cpp
```diff
@@ -7,48 +7,47 @@ void DataBrowser::userLeave(const std::string &userId)
...
-    if (it->second == nullptr)
+    auto function = []( IDataSelector* selector, std::vector<size_t>& returnValues ) -> bool
     {
-        return false;
-    }
-    return it->second->getDataType2(returnValues);
+        return selector->getDataType2( returnValues );
+    };
+
+    return safeCall( userId, [this, &returnValues, function]( const std::unique_ptr<IDataSelector>& selector ) -> bool
```
smay1613, line 29: here too

functors task (2)/student/include/impl.h
JohnDou1991: It was not compiling otherwise.

smay1613: Please provide compilation log and compiler version

PR #38, opened 2020-09-25
Iterators task 1 - Sakun

AndreySakun:


iterators task (1)/student/include/impl.h
```diff
@@ -2,20 +2,77 @@
...
-    /** @todo Iterator traits*/
-    /** @todo Constructor with container*/
-    /** @todo Copy assignment operator*/
-    /** @todo Move assignment operator*/
+private:
+    Container& mContainer;
+public:
+    /** Iterator traits*/
+    using iterator_category = std::output_iterator_tag;
+    using value_type =  typename Container::value_type;
```
smay1613, line 12: void is usually used for output iterators here
will your solution work with c arrays?

iterators task (1)/student/include/impl.h
```diff
@@ -2,20 +2,77 @@
...
+    using difference_type = typename Container::difference_type;
+    using pointer = typename Container::pointer;
+    using reference = typename Container::reference;
```
smay1613, line 15:
```suggestion
    using reference = void; 
```

https://stackoverflow.com/questions/2619551/output-iterators-value-type

iterators task (1)/student/include/impl.h
```diff
@@ -2,20 +2,77 @@
...
+
+    /** Move assignment operator*/
+    accumulator& operator=(value_type&& value)
+    {
+        mContainer.push_back(std::move(value));
+        return *this;
+    }
+
+    template <typename T>
+    accumulator& operator=(T&& value)
```
smay1613, line 49: please use more descriptive name, what is value?

smay1613: also, what is T?

iterators task (1)/student/include/impl.h
```diff
@@ -2,20 +2,77 @@
...
+    {
+        for (auto && item: value)
```
smay1613, line 51:
```suggestion
        std::move(value.begin(), value.end(), std::back_inserter(m_container));
```

iterators task (1)/student/include/impl.h
```diff
@@ -2,20 +2,77 @@
...
+    accumulator& operator=(const value_type& value)
+    {
+	    mContainer.push_back(value);
+	    return *this;
+    }
+
+    template <typename T>
+    accumulator& operator=(const T& value)
+    {
+        for (auto && item: value)
```
smay1613, line 34:
```suggestion
         m_container.insert(m_container.end(), value.begin(), value.end());
```

smay1613: When you have to write loop, always check available algorithms before!

iterators task (1)/student/include/impl.h
```diff
@@ -2,20 +2,77 @@
...
+        {
+            mContainer.push_back(item);
+        }
+        return *this;
+    }
+
+    /** Move assignment operator*/
+    accumulator& operator=(value_type&& value)
```
smay1613, line 42: is it required for task?

iterators task (1)/student/include/impl.h
```diff
@@ -2,20 +2,77 @@
...
+    /** Constructor with container*/
+    explicit accumulator(Container & container)
+    : mContainer (container)
+    {
 
-    /** @todo Iterator operations */
+    }
+
+    /** Copy assignment operator*/  
+    accumulator& operator=(const value_type& value)
```
smay1613, line 25: do we need this function?

smay1613 (changes requested):

iterators task (1)/student/include/impl.h
AndreySakun: Changed to void.

->will your solution work with c arrays?
Even after change it will not work with c arrays as they do not have neither insert nor push_back methods. Container there  is a type of object where we put data.

iterators task (1)/student/include/impl.h
AndreySakun: Changed to void also for difference_type and pointer

iterators task (1)/student/include/impl.h
AndreySakun: Agree. Changed.

iterators task (1)/student/include/impl.h
AndreySakun: No. Removed

iterators task (1)/student/include/impl.h
AndreySakun: removed

iterators task (1)/student/include/impl.h
AndreySakun: This looks much better :-)

iterators task (1)/student/include/impl.h
smay1613: agree

PR #41, opened 2020-09-26
Iterators task1 solution. Mykola Kobets.

kolya-kobets:


iterators task (1)/student/include/impl.h
```diff
@@ -5,11 +5,51 @@ template <class Container>
...
+        return *this;
+    }
+    accumulator<Container>& operator*()
+    {
+        return *this;
+    }
+
+    ///Mixin interface for an iterator & reference type is not
+    ///a canonical way to implement iterators.
+    ///Nevertheless it is done so for the sake of simplicity.
```
smay1613, line 35: Please provide more info, do you have any suggestion for this task?

iterators task (1)/student/include/impl.h
```diff
@@ -5,11 +5,51 @@ template <class Container>
 struct accumulator
 {
     /** @todo Iterator traits*/
+    using iterator_category = typename std::output_iterator_tag;
+    using value_type        = Container;
```
smay1613, line 9: does it has sense?
for output iterator it's traits are usually declared as void

iterators task (1)/student/include/impl.h
```diff
@@ -5,11 +5,51 @@ template <class Container>
...
+    }
+
+    template<typename C2>
+    accumulator<Container>& operator=(const C2& c2)
+    {
+        mContainer->insert(mContainer->end(), c2.begin(), c2.end());
+        return *this;
+    }
+private:
+    Container* mContainer = nullptr;
```
smay1613, line 52: Do you have any reason to declare this as pointer? Why not as reference?

kolya-kobets: yes, ref_wrapper a better approach here.

iterators task (1)/student/include/impl.h
kolya-kobets: that's right.

iterators task (1)/student/include/impl.h
kolya-kobets: it looks odd that operator* returns pointer to itself, but not to the reference:
it could be either T& or some adapter(ie map,vec<bool>)

kolya-kobets: in any case it would put assignment  operators to a more appropriate context.

iterators task (1)/student/include/impl.h
smay1613: Why not just plain reference? Client is responsible for maintaining iterators validity

iterators task (1)/student/include/impl.h
smay1613: Still can't understand, could you please highlight the problem?
operator* should return dereferenced pointer of _this_, so return type is reference

iterators task (1)/student/include/impl.h
kolya-kobets: reference will make copy assignment operator invalid.

iterators task (1)/student/include/impl.h
kolya-kobets: even the current iface fulfills requirements to iterator. it is inconsistent. ie. operator* is redundant, as we can assign c2 directly to iterator. basically it breaks single responsibility principle.

iterators task (1)/student/include/impl.h
smay1613: Agree, it's a good point, thanks

PR #48, opened 2020-09-27
functors task (2) - Pozdniakov

AlexPozdniakov:


functors task (2)/student/include/impl.h
```diff
@@ -53,31 +53,39 @@ class DataBrowser
...
-        // call functor
-        return false;
-    }
+    bool safeCall(const std::string &userId, T&& f) const;
 
     /**
-     * @todo You should implement this method, that is responsible for a "general" data requesting
-     * General data request will return bool, accept the selector and output param
-     * You must adapt functors before invocation
+     * General data requesting via a single interface.
```
smay1613, line 63: Is it a suggestion for the task description improvement? If it is, please create pull request

AlexPozdniakov: No, it's just something like documentation after I have implemented the functional. It helps me remember technical details. It is not always clear, for example, which signature is appropriate for "template" functionals params.
As a variant, I would rather set the type of functional param a bit explicitly, for example, 
```
template<typename Output>
bool invokeDataRequest(
  std::function<bool(IDataSelector*, Output)> method,
  const std::unique_ptr<IDataSelector>& selector,
  Output& result) const
```
but for study purposes that is nice as is :) really challenging task.

PR #51, opened 2020-09-28
Iterators task 1 - Poplavskyi

NetLagina:


iterators task (1)/student/include/impl.h
```diff
@@ -4,12 +4,41 @@
 template <class Container>
 struct accumulator
 {
+private:
+    Container* m_container = nullptr;
```
smay1613, line 8: do you have any reason to declare it as pointer?

iterators task (1)/student/include/impl.h
```diff
@@ -4,12 +4,41 @@
...
+public:
     /** @todo Iterator traits*/
+    using iterator_category = std::output_iterator_tag;
+    using value_type = Container;
```
smay1613, line 12: does it have sense?
for output iterators, traits are usually defined as void

iterators task (1)/student/include/impl.h
```diff
@@ -4,12 +4,41 @@
...
+    using difference_type = std::ptrdiff_t;
+    using pointer = Container*;
+    using reference = Container&;
+
     /** @todo Constructor with container*/
+    accumulator<Container>(Container& container)
```
smay1613, line 18: prefer compiler-generated versions

smay1613 (changes requested):

PR #53, opened 2020-09-28
iterator task #1 tdobrovolska

TetianaDobrovolska:


iterators task (1)/student/include/impl.h
```diff
@@ -5,11 +5,46 @@ template <class Container>
 struct accumulator
 {
     /** @todo Iterator traits*/
+    using traits = std::iterator_traits<typename Container::iterator>;
+    using difference_type = typename traits::difference_type;
```
smay1613, line 9: does it has sense?
for output iterators, all properties are usually defined as void

TetianaDobrovolska: correct

PR #55, opened 2020-09-28
algorithm task 2 VNevzorov

sam-mcdonalds:


smay1613 (dismissed):
upd: see comment below

algorithms task (2)/student/include/impl.h
```diff
@@ -9,4 +10,6 @@
  *
  * @return true if source starts from a prefix
  */
-bool startsWith(const std::string& source, const std::string& prefix);
+bool startsWith(const std::string& source, const std::string& prefix){
+  return std::equal(prefix.begin(), prefix.end(), source.begin());
```
smay1613, line 14: what if source is less than prefix?

PR #56, opened 2020-09-28
algorithms task (2) - Pozdniakov

AlexPozdniakov:


algorithms task (2)/student/src/impl.cpp
```diff
@@ -1 +1,6 @@
 #include "impl.h"
+
+bool startsWith(const std::string& source, const std::string& prefix)
+{
+    return std::equal(prefix.begin(), prefix.end(), source.begin());
```
smay1613, line 5: what if source less than prefix?

PR #60, opened 2020-09-28
algorithms_task_2_rudnichenko

CatBehemotth:
Another ways:
1. Use string::substr
2. Character-by-character comparison


algorithms task (2)/student/src/impl.cpp
```diff
@@ -1 +1,6 @@
 #include "impl.h"
+
+bool startsWith(const std::string& source, const std::string& prefix)
+{
+    return std::equal(source.begin(),source.begin()+prefix.size(),prefix.begin());
```
smay1613, line 5: what if source is less than prefix?

PR #61, opened 2020-09-28
Algorithms task 2 - Poplavskyi

NetLagina:
another good variant for C++20 is

```c++
return source.starts_with(prefix);
```

### DIFF
--- a/algorithms task (2)/student/src/impl.cpp
+++ b/algorithms task (2)/student/src/impl.cpp
@@ -1,1 +1,6 @@
 #include "impl.h"
+
+bool startsWith(const std::string& source, const std::string& prefix)
+{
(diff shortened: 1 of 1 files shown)


algorithms task (2)/student/src/impl.cpp
```diff
@@ -1 +1,6 @@
 #include "impl.h"
+
+bool startsWith(const std::string& source, const std::string& prefix)
+{
+	return std::equal(prefix.begin(), prefix.end(), source.begin());
```
smay1613, line 5: what if source is less than prefix?

NetLagina: Added size check

PR #64, opened 2020-09-28
algorithms task 2 - Artur Hnoinskyi

ArchiGn98:


algorithms task (2)/student/src/impl.cpp
```diff
@@ -1 +1,4 @@
 #include "impl.h"
+bool startsWith(const std::string& source, const std::string& prefix){
+    return prefix.compare(source.substr(0, prefix.size())) == 0 ? true : false;
```
smay1613, line 3: prefer algorithm version

PR #72, opened 2020-09-28
Algorithms t3  MKobets

kolya-kobets:


algorithms task (3)/student/include/impl.h
```diff
@@ -13,4 +15,14 @@
  * @note behavior is undefined if shift out of bounds
  */
 template<class BiIt>
-BiIt shift_right_n(BiIt begin, BiIt end, typename BiIt::difference_type shift);
+BiIt shift_right_n(BiIt begin, BiIt end, typename BiIt::difference_type shift)
+{
+    ///Didn't get the idea for "end" parameter, as shift specifies the number of elements to shift.
```
smay1613, line 20: Task will be changed, thanks

algorithms task (3)/student/include/impl.h
```diff
@@ -13,4 +15,14 @@
...
+    ///begin+shift - start of the destination sequence
+    //return std::move_backward(begin, std::next(begin, shift), std::next(begin, shift*2));
+
+    //source & dest ranges never overlap, so move can be used instead of move_backward.
```
smay1613, line 24: Why are you sure that source and destination will not overlap?

kolya-kobets: shift - specifies the number of elements to move
begin - source
begin+shift - destination

smay1613: range:
1   2   3   4   5
^   .   ^ 
b   .    e

n = 1

result:
1 1 2 4 5

don't they overlap?

kolya-kobets: as far as I understood, end is ignored for the task. range to move is specified by  (begin, split) pair.

smay1613: we shift the subrange

PR #77, opened 2020-09-28
Functors t2  MKobets

kolya-kobets:


functors task (2)/student/include/impl.h
```diff
@@ -76,8 +81,6 @@ class DataBrowser
     template<class Functional, typename Output>
     bool invokeDataRequest(Functional method, const std::unique_ptr<IDataSelector>& selector, Output& result) const
     {
-        // adapt function
-        // call selector member
-        return false;
+        return std::bind(method, selector.get(), std::ref(result))();
```
smay1613, line 84: why not std::function?

functors task (2)/student/src/impl.cpp
```diff
@@ -5,50 +5,56 @@ void DataBrowser::userLeave(const std::string &userId)
...
-    {
-        return false;
-    }
-    if (it->second == nullptr)
-    {
-        return false;
-    }
-    return it->second->getDataType1(returnValues, 0);
+    return safeCall(userId,
+        [this, &returnValues](const std::unique_ptr<IDataSelector>& selector)
```
smay1613, line 16: please give this lambda a name

PR #83, opened 2020-09-29
algorithms task #3 TDobrovolska

TetianaDobrovolska:


algorithms task (3)/student/include/impl.h
```diff
@@ -13,4 +13,8 @@
  * @note behavior is undefined if shift out of bounds
  */
 template<class BiIt>
-BiIt shift_right_n(BiIt begin, BiIt end, typename BiIt::difference_type shift);
+BiIt shift_right_n(BiIt begin, BiIt end, typename BiIt::difference_type shift)
+{
+    std::move_backward(begin, end, std::next(end, shift));
+    return std::next(begin, shift);
```
smay1613, line 19: move_backward should return this iterator

TetianaDobrovolska: correct

PR #84, opened 2020-09-29
Implement iterators task #2 - ENikonov

re-nikonov:


iterators task (2)/student/include/impl.h
```diff
@@ -1,33 +1,53 @@
...
 template<class Iterator, class T>
 struct pair_iterator<Iterator, false, T> : public Iterator
 {
-    /** @todo Iterator traits */
-    /** @todo Iterator operations */
-    /** @todo Constructor from original iterator */
+    using value_type = typename T::second_type;
+    using reference =
+        typename std::conditional<
+            std::is_const<typename std::remove_reference<typename std::iterator_traits<Iterator>::reference>::type>::value,
```
smay1613, line 39: Nice!
Consider is_const trait extraction to a named struct

iterators task (2)/student/include/impl.h
```diff
@@ -1,33 +1,53 @@
...
+            const value_type&,
+            value_type&
+        >::type;
+    using pointer = typename std::add_pointer<reference>::type;
+
+    explicit pair_iterator(Iterator iterator) : Iterator(std::move(iterator)) {}
```
smay1613, line 45: why do you need move here?

re-nikonov: I don't. Just used to move everything what isn't lvalue reference. But this is just an iterator... removed std::move().

iterators task (2)/student/include/impl.h
re-nikonov: Done! But the merge request has been closed... Anyway it was interesting!

PR #89, opened 2020-09-29
Implement algorithms task #1 - ENikonov

re-nikonov:


algorithms task (1)/student/include/impl.h
```diff
@@ -11,4 +13,15 @@
...
+std::pair<FI, FI> consecutive_group(FI start, FI end, Comparator comp = {})
+{
+    const auto startIt = std::adjacent_find(start, end, comp);
+    if (startIt == end)
+    {
+        return std::make_pair(end, end);
+    }
+    const auto endIt = std::find_if_not(std::next(startIt, 2),
+                                        end,
+                                        std::bind(comp, *startIt, std::placeholders::_1));
```
smay1613, line 25: why passed as not ref?

re-nikonov: My bad... Fixed.

PR #92, opened 2020-09-29
Implement algorithms task #4 - ENikonov [new unit tests... again, task statement changes]

re-nikonov:
- add unit tests

### DIFF
--- a/algorithms task (4)/student/include/impl.h
+++ b/algorithms task (4)/student/include/impl.h
@@ -1,14 +1,27 @@
 #pragma once
+
 #include <algorithm>
+#include <utility>
 
 /**
  * @todo Implement sliding algorithm (left/right shift of subsequence to a position)
  *
  * @param begin - begin of random-access range
  * @param last - end of random-access range
- * @param pos - new begin of slided range
+ * @param pos - new begin of slided range to the left, or new end of slided range to the right.
  *
  * @return new slided range position
  */
 template <typename RandomAccessIterator>
-std::pair<RandomAccessIterator, RandomAccessIterator> slide(RandomAccessIterator begin, RandomAccessIterator last, RandomAccessIterator pos);
+std::pair<RandomAccessIterator, RandomAccessIterator> slide(RandomAccessIterator begin, RandomAccessIterator last, RandomAccessIterator pos)
+{
+    if (pos < begin) // left shift
+    {
+        return {pos, std::rotate(pos, begin, last)};
+    }
+    if (last < pos) // right shift
+    {
+        return {std::rotate(begin, last, pos), pos};
+    }
+    return {begin, last};
+}
--- a/algorithms task (4)/student/tests/unit_tests/src/maintest.cpp
+++ b/algorithms task (4)/student/tests/unit_tests/src/maintest.cpp
@@ -6,11 +6,11 @@ using namespace ::testing;
 TEST(SlideSame, Right)
 {
     std::deque<size_t> data {1, 1, 1, 0, 0, 0, 0};
-
     std::deque<size_t> expected {0, 0, 0, 0, 1, 1, 1};
 
     constexpr size_t offset {3};
     auto endOfSlidingRange = std::next(data.begin(), offset);
+
     auto newSlidedRange = slide(data.begin(), endOfSlidingRange,
                                 data.end());
 
@@ -20,7 +20,7 @@ TEST(SlideSame, Right)
     EXPECT_EQ(newSlidedRange.second, data.end());
 }
 
-TEST(SlideSame, Left)
+TEST(SlideSame, ReverseLeft)
 {
     std::deque<size_t> data {0, 0, 0, 0, 1, 1, 1};
     std::deque<size_t> expected {1, 1, 1, 0, 0, 0, 0};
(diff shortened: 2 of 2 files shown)


algorithms task (4)/student/include/impl.h
```diff
@@ -1,14 +1,38 @@
...
  * @return new slided range position
  */
 template <typename RandomAccessIterator>
-std::pair<RandomAccessIterator, RandomAccessIterator> slide(RandomAccessIterator begin, RandomAccessIterator last, RandomAccessIterator pos);
+std::pair<RandomAccessIterator, RandomAccessIterator> slide(RandomAccessIterator begin, RandomAccessIterator end, RandomAccessIterator newEnd)
+{
+    const auto rangeSize = std::distance(begin, end);
+    const auto newBegin = std::prev(newEnd, rangeSize);
+
+    std::vector<typename std::iterator_traits<RandomAccessIterator>::value_type> tmp;
```
smay1613, line 23: 1. algorithms usually use get_temporary_buffer (it was designed for it)
2. consider renaming

algorithms task (4)/student/include/impl.h
```diff
@@ -1,14 +1,38 @@
...
+    tmp.reserve(rangeSize);
+    std::move(begin, end, std::back_inserter(tmp));
```
smay1613, line 25: do you really need this?

algorithms task (4)/student/include/impl.h
```diff
@@ -1,14 +1,38 @@
...
+#include <vector>
 
 /**
  * @todo Implement sliding algorithm (left/right shift of subsequence to a position)
  *
  * @param begin - begin of random-access range
- * @param last - end of random-access range
- * @param pos - new begin of slided range
+ * @param end - end of random-access range
+ * @param newEnd - new end of slided range
```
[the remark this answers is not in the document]
re-nikonov: It is from unit tests: otherwise unit tests fail if implement as new begin.
... or unit tests should be fixed?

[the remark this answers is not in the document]
smay1613: tests are right, please consider std::rotate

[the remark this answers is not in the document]
re-nikonov: True, tests are right! Didn't understand the task... or understood it incorrectly. Now it's done.

algorithms task (4)/student/include/impl.h
re-nikonov: 1. Got it!
2. Nothing to rename :)

algorithms task (4)/student/include/impl.h
```diff
@@ -11,4 +13,7 @@
  * @return new slided range position
  */
 template <typename RandomAccessIterator>
-std::pair<RandomAccessIterator, RandomAccessIterator> slide(RandomAccessIterator begin, RandomAccessIterator last, RandomAccessIterator pos);
+std::pair<RandomAccessIterator, RandomAccessIterator> slide(RandomAccessIterator begin, RandomAccessIterator last, RandomAccessIterator pos)
+{
+    return std::make_pair(std::rotate(begin, last, pos), pos);
```
smay1613, line 18: _there can be more effective version with less amount of swaps :)_

algorithms task (4)/student/include/impl.h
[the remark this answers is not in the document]
re-nikonov: Now I don't understand again!
I believe that "pos" and "last" parameters are not in their correct place, I mean they should be swapped. And if using of std::rotate() is correct solution for this task, this is proof of my "theory".
Consider for example the first 2 test cases of SlideSame: data.[r]end() is passed as "pos" parameter, but HOW data.[r]end() can be "new begin of slided range"??
This is very confusing, that is why first I decided that "pos" parameter is actually "new END of slided range".
Looks like it should be fixed...

algorithms task (4)/student/include/impl.h
re-nikonov: Hmmm...
As I understand it right the most effective std::rotate() implementation does N (or N-1, not so important) swaps, i.e. 3*N moves. Well, it can be done in much less moves, N + min(n1, n2), where n1, n2 are lengths of subranges, but it will require allocation of additional memory to hold temporary min(n1, n2) elements, it is almost the same I did in the first try. But we all know that allocating new memory is always heavy operation and most likely it will work slower than just N swaps in place.
What exactly did you mean by "there can be more effective version..." ??

smay1613: please check proposed solution (check your email)
just make this rotate conditionally more effective, don't move the whole range

re-nikonov: No, wait.

I gave up, I checked the solution. Now I see. But as I said, the task's statement (in "impl.h") is very confusing.
 * @param pos - new begin of slided range.
Here "pos" is new BEGIN of slided range in case of left shift, it is Ok and clear, BUT it is new END in case of right shift, and here is the reason of confusing. It would be clear if:
 * @param pos - new begin of slided range to the left, or new end of slided range to the right.
We all know that end is not begin... :)

More of that, unit tests, they test ONLY right shift case:
    if (last < pos) // right shift
    {
        return {std::rotate(begin, last, pos), pos};
    }
In case of left shift my solution will definitely crash (not acceptable). As well as everyone other's solution since everybody submitted the SAME solution (NO accepted solutions at all). What I want to say is that condition:
    if (pos < begin) // left shift
    {
        return {pos, std::rotate(pos, begin, last)};
    }
has never executed, because no unit tests for that case at all. And if it had tests, then nobody submitted current solution as it would crash and students would continue finding the right solution. It must be fixed.

See my changes in new update of the merge request. And with these new unit tests all solutions will crash (except your's of course).

re-nikonov:
Is this task accepted? Shall we consider this task as finished?

algorithms task (4)/student/include/impl.h
smay1613: Oh, thanks, you're right. I will fix the tests, thanks!

smay1613:
Thanks for the contribution!

smay1613:
@re-nikonov thank you again, task is fixed now:
https://github.com/smay1613/LTC-STL-Student/pull/259

PR #96, opened 2020-10-01
algorithms task (6) - Sakun

AndreySakun:


algorithms task (6)/student/include/impl.h
```diff
@@ -19,4 +19,11 @@
  */
 
 template <typename Sequence, typename FwIt>
-void insert_to_sorted(Sequence& target, const FwIt sourceBegin, const FwIt sourceEnd);
+void insert_to_sorted(Sequence& target, const FwIt sourceBegin, const FwIt sourceEnd)
+{
+    for (FwIt it = sourceBegin; it != sourceEnd; ++it)
+    {
+        auto insertPosition = std::upper_bound(begin(target), end(target), *it);
```
smay1613, line 26: will it be effective if our range will consist of duplicated values?

PR #98, opened 2020-10-01
Implement algorithms task #5 - ENikonov [new unit tests]

re-nikonov:


algorithms task (5)/student/include/impl.h
```diff
@@ -10,4 +13,12 @@
  * @return gathered range
  */
 template <typename BiIt, typename UnaryPredicate>
-std::pair<BiIt, BiIt> gather(BiIt begin, BiIt end, BiIt position, UnaryPredicate predicate);
+std::pair<BiIt, BiIt> gather(BiIt begin, BiIt end, BiIt position, UnaryPredicate predicate)
+{
+    const auto partitioningEnd = std::partition(begin, end, predicate);
+    const auto gatheringOffset = std::min(std::distance(begin, position),
```
smay1613, line 19: distance call + rotate is not effective operation here, please consider  simpler algorithm

re-nikonov: Found and implemented new algorithm.
Please take in account my new unit tests condition about that "right" element must be at specified position.

algorithms task (5)/student/include/impl.h
```diff
@@ -10,4 +13,19 @@
...
+    {
+        return !predicate(value);
+    };
+
+    auto partitionBegin = std::partition(begin, std::next(position), notPredicate);
+    auto partitionEnd = std::partition(position, end, predicate);
+
+    if (partitionBegin == std::next(position))
+    {
+        std::advance(partitionBegin, -1);
```
smay1613, line 28: What if partition begin is at "begin"?

algorithms task (5)/student/include/impl.h
```diff
@@ -10,4 +13,19 @@
...
+{
+    const auto notPredicate = [predicate](const typename std::iterator_traits<BiIt>::reference value)
+    {
+        return !predicate(value);
+    };
+
+    auto partitionBegin = std::partition(begin, std::next(position), notPredicate);
+    auto partitionEnd = std::partition(position, end, predicate);
+
+    if (partitionBegin == std::next(position))
```
smay1613, line 26: What if position is end?

algorithms task (5)/student/include/impl.h
re-nikonov: Partition's begin cannot be "begin" here as it equals to std::next(position), so it will be at least "begin" + 1.

algorithms task (5)/student/include/impl.h
re-nikonov: Undefined behavior, but I assumed that it doesn't make sense to pass end() as position. I think it is not so important here. Do you think it should be fixed?

smay1613: consider the empty range

algorithms task (5)/student/include/impl.h
smay1613: please add tests for empty range

re-nikonov: Right! Nothing is added, the test for empty range already exists. But the algorithm is changed.

algorithms task (5)/student/include/impl.h
re-nikonov: Fixed.

re-nikonov:
Is this task accepted? Shall we consider this task as finished?

smay1613:
Everything is fine here now

PR #99, opened 2020-10-01
Implement algorithms task #6 - ENikonov

re-nikonov:


algorithms task (6)/student/include/impl.h
```diff
@@ -19,4 +21,16 @@
  */
 
 template <typename Sequence, typename FwIt>
-void insert_to_sorted(Sequence& target, const FwIt sourceBegin, const FwIt sourceEnd);
+void insert_to_sorted(Sequence& target, const FwIt sourceBegin, const FwIt sourceEnd)
+{
+    static_assert(std::is_const<typename std::remove_reference<
```
smay1613, line 26: thanks, will merge this later :)

algorithms task (6)/student/include/impl.h
```diff
@@ -19,4 +21,16 @@
...
+                      typename std::iterator_traits<FwIt>::reference>::type>::value,
+                  "FwIt must point to const data");
+
+    // Use std::upper_bound() for stable order of source values.
+
+    std::for_each(sourceBegin, sourceEnd, [&target](typename std::iterator_traits<FwIt>::reference sourceValue)
```
smay1613, line 32: please avoid for_each

algorithms task (6)/student/include/impl.h
```diff
@@ -19,4 +21,16 @@
...
 
 template <typename Sequence, typename FwIt>
-void insert_to_sorted(Sequence& target, const FwIt sourceBegin, const FwIt sourceEnd);
+void insert_to_sorted(Sequence& target, const FwIt sourceBegin, const FwIt sourceEnd)
+{
+    static_assert(std::is_const<typename std::remove_reference<
+                      typename std::iterator_traits<FwIt>::reference>::type>::value,
+                  "FwIt must point to const data");
+
+    // Use std::upper_bound() for stable order of source values.
```
smay1613, line 30: but there is another trade-off
will your solution be effective in case if range will consist of duplicated values?

algorithms task (6)/student/include/impl.h
smay1613: but here I think it is acceptable

re-nikonov: I believe this is the only reason why std::for_each() is still alive, because for (auto it = begin; it != end; ++it) doesn't look better at all than std::for_each(begin, end, lambda);

algorithms task (6)/student/include/impl.h
re-nikonov: I don't understand what I should do in this case... Don't insert duplicated values at all? Or what? Give me a clue.

smay1613: lower bound

algorithms task (6)/student/include/impl.h
smay1613: you always should prefer range for, but here you have only iterators, so it is ok

smay1613, line 32: why not const reference?

re-nikonov: Because we've already tested by static_assert() that reference is constant.
Anyway changed to "const value_type&" as it is clearer to understand, what's your reckon?

algorithms task (6)/student/include/impl.h
re-nikonov: Wait a second!
In case of lower_bound() it will be "reversed stable" solution. Because every new duplicate source element will be added BEFORE previous duplicate source element. It's not definition of "stable", as final sequence will have source's duplicate elements in reverse order. Won't it?

smay1613: We must satisfy only "sorted" requirement, there is no unit tests for "stable sorted" requirement

re-nikonov: Then why is there question about stability in task description?
@warning IS YOUR SOLUTION STABLE?
In case of lower_bound() the answer will be "No", that makes this question weird.
Further, if we ignore this "stable" question and will use lower_bound() instead of upper_bound(), I don't see any profit: there will be O(log(N)) comparisons in both cases and (possibly I assume) a few more iterator increments in case of upper_bound(), but no, actually it will be O(N) increments in both cases. Am I right?
So how lower_bound() solution can be more effective? Am I missing something?

smay1613: This question is just a hint to think :)
Please consider such corner case, when all values are duplicated.
Example:
Sorted: 1 1 1 1 1 1 1 1 1 1
Insert: 1
Result (lower bound): **1** 1 1 1 1 1 1 1 1 1 1 (O(1) here) 
Result (upper bound): 1 1 1 1 1 1 1 1 1 1 **1** (O(n))

In case of upper_bound, we need N additional operations, where N is the equal_range size.

re-nikonov: Which additional operations do you mean?

lower_bound() as well as upper_bound() are binary search algorithms: there will be O(log(N)) comparisons in both cases even on equal ranges (it is still binary search on equal ranges), specifically in this case 4 and 3 comparisons respectively. Neither 1 nor N. Logarithm is guaranteed by the standard! If we have N operations instead of log(N) - then this is not *standard* template library.

Please check GCC implementation:
https://github.com/gcc-mirror/gcc/blob/d9375e490072d1aae73a93949aa158fcd2a27018/libstdc%2B%2B-v3/include/bits/stl_algobase.h#L1023
https://github.com/gcc-mirror/gcc/blob/d9375e490072d1aae73a93949aa158fcd2a27018/libstdc%2B%2B-v3/include/bits/stl_algo.h#L2028

LLVM implementation is the same.

smay1613: You are right, algorithm is still binary search.
But lower bound is more effective, because we don't need to find greater element, or iterate to the end.
In this task it is not required to use upper bound.
Please add a new test with vector of 1 000 000 duplicated values (as range, where we will insert) and you should see the difference in practice. Inserted range should contain the same duplicated value. 
Lower bound will use just one comparison. 
Upper bound will use 1 000 000 comparisons. 
Consider this corner case.

re-nikonov: Created unit test with std::list (or std::vector, doesn't matter) of 1'000'000 duplicated values. With counters of iterator increments and value comparisons.
Ran std::lower_bound() and std::upper_bound() to see the difference in increments and comparisons.
Tested with GCC and Clang.
Please see results.

smay1613: You are right, sorry for this.
std::distance is called inside, because of ForwardIterator requirement:
https://gcc.gnu.org/onlinedocs/libstdc++/libstdc++-html-USERS-4.0/stl__algo_8h-source.html#l02624

smay1613: I'll fix the comment later

PR #127, opened 2020-12-11
Sequence task 2 Kolesnichenko Oleksandr

Kolesni44enko:


sequence_task_2/student/src/impl.cpp
```diff
@@ -1 +1,9 @@
 #include "impl.h"
+
+template<class T, class BI>
+void rotate(std::list<T>& list, BI begin, BI new_begin, BI end) {
+    list.splice(begin, list, new_begin, end);
+}
+
+template void rotate(std::list<size_t>&, std::list<size_t>::iterator,
```
smay1613, line 8: Use explicit instantiation in .cpp unit

smay1613:
Nice

PR #150, opened 2020-12-13
sequence task #3 TDobrovolska

TetianaDobrovolska:


sequence_task_3/student/src/impl.cpp
```diff
@@ -1,2 +1,20 @@
 #include "impl.h"
 
+std::string reverseWords(const std::string& input)
+{
+    std::istringstream input_stream(input);
+    std::deque<std::string> words;
+    std::move(std::istream_iterator<std::string>(input_stream),
```
smay1613, line 7: Nice try, but when dereferenced, const lvalue reference is returned for istream iterator:
https://en.cppreference.com/w/cpp/iterator/istream_iterator/operator*

sequence_task_3/student/src/impl.cpp
```diff
@@ -1,2 +1,20 @@
...
+              std::istream_iterator<std::string>(),
+              std::front_inserter(words));
+
+    std::ostringstream output_stream;
+    std::move(words.begin(), words.end(),
```
smay1613, line 12: Won't work too :(
https://en.cppreference.com/w/cpp/iterator/ostream_iterator/operator%3D

PR #155, opened 2020-12-13
sequence_task_3 amolodchenkova

nussa909:


sequence_task_3/student/src/impl.cpp
```diff
@@ -1,2 +1,26 @@
...
+#include <algorithm>
+
+std::string reverseWords(const std::string& input)
+{
+    std::stringstream istream(input);
+    std::deque<std::string> deque;
+
+    std::for_each(std::istream_iterator<std::string>(istream), std::istream_iterator<std::string>(), [&deque]( const std::string& s )
+    {
+        deque.push_front(s);
```
smay1613, line 11: _Prefer std::front_inserter_

sequence_task_3/student/src/impl.cpp
```diff
@@ -1,2 +1,26 @@
 #include "impl.h"
+#include <algorithm>
+
+std::string reverseWords(const std::string& input)
+{
+    std::stringstream istream(input);
+    std::deque<std::string> deque;
+
+    std::for_each(std::istream_iterator<std::string>(istream), std::istream_iterator<std::string>(), [&deque]( const std::string& s )
```
smay1613, line 9: Prefer more appropriate algorithm rather than std::for_each.
For example, std::copy is much suitable here.

sequence_task_3/student/src/impl.cpp
```diff
@@ -1,2 +1,26 @@
...
+    {
+        deque.push_front(s);
+    });
+
+    std::stringstream outstream;
+    std::for_each(deque.begin(), deque.end(), [&outstream, &deque]( const std::string& word )
```
smay1613, line 15: use output stream iterators

sequence_task_3/student/src/impl.cpp
```diff
@@ -1,2 +1,26 @@
...
+    {
+        outstream << word;
+        if (word != deque.back())
```
smay1613, line 18: add a test case with duplicated words and see how string equality operator will work here :)

smay1613 (changes requested):

sequence_task_3/student/src/impl.cpp
```diff
@@ -4,23 +4,17 @@
 std::string reverseWords(const std::string& input)
 {
     std::stringstream istream(input);
-    std::deque<std::string> deque;
-
-    std::for_each(std::istream_iterator<std::string>(istream), std::istream_iterator<std::string>(), [&deque]( const std::string& s )
-    {
-        deque.push_front(s);
-    });
+    std::deque<std::string> deque { std::istream_iterator<std::string>(istream), std::istream_iterator<std::string>() };
```
smay1613, line 7: why not vector?

PR #158, opened 2020-12-14
Adaptors task Oleksandr Stepanenko

stepanenkoalvl:


adaptors task/student/src/impl.cpp
```diff
@@ -1,8 +1,32 @@
...
     {'{', '}'}
 };
+
+bool isValid(const std::string& source)
+{
+    std::stack<char> stackChecker;
+
+    for (auto nextSymbol: source) {
+        for (auto nextPair: brackets) {
+            if (nextSymbol == nextPair.first) {
```
smay1613, line 17: Please name this condition like
const bool isOpenBracket = nextSymbol == nextPair.first;

adaptors task/student/src/impl.cpp
```diff
@@ -1,8 +1,32 @@
...
+                stackChecker.push(nextSymbol);
+            }
+            if (nextSymbol == nextPair.second) {
```
smay1613, line 20: isClosedBracket

adaptors task/student/src/impl.cpp
```diff
@@ -1,8 +1,32 @@
...
+                if (!stackChecker.empty() && stackChecker.top() == nextPair.first){
+                    stackChecker.pop();
+                } else {
+                    stackChecker.push(nextSymbol);
+                }
+            }
+        }
+    }
+
+    if (stackChecker.empty()){ return true; }
```
smay1613, line 30:
```suggestion
    return stackChecker.empty()
```

adaptors task/student/src/impl.cpp
```diff
@@ -1,8 +1,32 @@
...
+    for (auto nextSymbol: source) {
+        for (auto nextPair: brackets) {
+            if (nextSymbol == nextPair.first) {
+                stackChecker.push(nextSymbol);
+            }
+            if (nextSymbol == nextPair.second) {
+                if (!stackChecker.empty() && stackChecker.top() == nextPair.first){
+                    stackChecker.pop();
+                } else {
+                    stackChecker.push(nextSymbol);
```
smay1613, line 24: Shouldn't we break here?
For example, what if we see the closed bracket when stack is empty

smay1613 (changes requested):

PR #163, opened 2020-12-14
Adaptors task gulov

GIurii:


adaptors task/student/include/impl.h
```diff
@@ -11,4 +13,48 @@
...
+         else
+         {
+            return false;
+         }
+      }};
+
+   for (const char& ch : source)
+   {
+      switch (ch) {
+      case '(':
```
smay1613, line 36: USE "brackets" VECTOR FROM .cpp FILE! It is mandatory!

smay1613 (changes requested):

adaptors task/student/include/impl.h
GIurii: Already reworked. Please check.

adaptors task/student/src/impl.cpp
```diff
@@ -1,8 +1,36 @@
...
+   for (const auto& ch : source)
+   {
+      for(const auto& pair: brackets){
+         if (ch == pair.first){
+            charsStack.push(pair.second);
+         } else if (ch == pair.second) {
+            if(!charsStack.empty() && pair.second == charsStack.top()) {
+               charsStack.pop();
+            } else {
+               result = false;
```
smay1613, line 25: return false is much simpler

adaptors task/student/src/impl.cpp
```diff
@@ -1,8 +1,36 @@
...
+{
+   std::stack<char> charsStack;
+   bool result = true;
+
+   for (const auto& ch : source)
+   {
+      for(const auto& pair: brackets){
+         if (ch == pair.first){
+            charsStack.push(pair.second);
+         } else if (ch == pair.second) {
```
smay1613, line 21: please give this condition a name like
const bool isClosedBracket =

smay1613 (changes requested):

PR #173, opened 2020-12-16
sequence_task_3 VRazvodovskyi

7e1ght:


sequence_task_3/student/src/impl.cpp
```diff
@@ -1,2 +1,18 @@
...
+std::string reverseWords(const std::string& input)
+{
+    std::istringstream iss(input);
+
+    auto res = std::accumulate(std::istream_iterator<std::string>(iss), 
+                               std::istream_iterator<std::string>(), 
+                               std::string(),
+                               [](std::string a, std::string b) { return b + " " + a; });
+    
+    if(true != res.empty()) res.pop_back();
```
smay1613, line 15: if (!res.empty()) will be simpler

sequence_task_3/student/src/impl.cpp
```diff
@@ -1,2 +1,18 @@
 #include "impl.h"
+#include <sstream>
+#include <iterator>
+#include <numeric>
 
+std::string reverseWords(const std::string& input)
+{
+    std::istringstream iss(input);
+
+    auto res = std::accumulate(std::istream_iterator<std::string>(iss),
```
smay1613, line 10: that's quite interesting, something new :)
please don't use std::accumulate for strings
https://stackoverflow.com/questions/15347123/how-to-construct-a-stdstring-from-a-stdvectorstring

sequence_task_3/student/src/impl.cpp
```diff
@@ -1,2 +1,18 @@
...
+#include <numeric>
 
+std::string reverseWords(const std::string& input)
+{
+    std::istringstream iss(input);
+
+    auto res = std::accumulate(std::istream_iterator<std::string>(iss), 
+                               std::istream_iterator<std::string>(), 
+                               std::string(),
+                               [](std::string a, std::string b) { return b + " " + a; });
```
smay1613, line 13: why not reference?
double copy

PR #182, opened 2020-12-16
unordered_assotiative_task_1_Stepanenko_Oleksandr

stepanenkoalvl:


unordered associative task (1)/student/src/impl.cpp
```diff
@@ -1 +1,22 @@
...
+
+    for (auto symbol: first) {
+        firstMap[symbol]++;
+    }
+
+    for (auto symbol: second) {
+        secondMap[symbol]++;
+    }
+
+    if (firstMap == secondMap) {
```
smay1613, line 17: _return firstMap == secondMap;_

PR #184, opened 2020-12-16
ordered_associative_task_Stepanenko_Oleksandr

stepanenkoalvl:


ordered associative task/student/src/impl.cpp
```diff
@@ -1 +1,14 @@
 #include "impl.h"
+
+size_t convertMarkTo5PointSystem(size_t markIn100Point)
+{
+    static std::map<size_t, size_t> marks {{30, 1}, {50, 2}, {70, 3}, {90, 4}};
+
+    for (auto markPair: marks) {
```
smay1613, line 7: wtf?
please use embedded search algorithms of associative container

ordered associative task/student/src/impl.cpp
```diff
@@ -1 +1,14 @@
...
+        if (markIn100Point < markPair.first) {
+            return markPair.second;
+        }
+    }
+
+    return 5; // + out of bound cases
```
smay1613, line 13: avoid this default value

smay1613 (changes requested):

PR #185, opened 2020-12-16
Unordered_associative_task_1_Gershkul

DevASG:


unordered_associative_task_1/student/src/impl.cpp
```diff
@@ -1 +1,9 @@
 #include "impl.h"
+#include "unordered_set"
+
+bool is_anagram(const std::string& first, const std::string& second)
+{
+    const std::unordered_multiset<char> set_f(first.begin(), first.end());
```
smay1613, line 6: In such case, ordered multiset will be much simpler.

Smart, but hard to understand and non traditional way :)

Not sure if it will work for complex objects (not the char), because order is unknown

PR #188, opened 2020-12-16
Ordered associative task gulov

GIurii:


ordered associative task/student/src/impl.cpp
```diff
@@ -1 +1,8 @@
 #include "impl.h"
+
+size_t convertMarkTo5PointSystem(size_t markIn100Point)
+{
+   std::map<size_t, size_t> ranges{{29, 1}, {49, 2}, {69, 3}, {89, 4}};
+   const auto it = ranges.lower_bound(markIn100Point);
+   return it == ranges.end() ? 5 : it->second;
```
smay1613, line 7: Please avoid the default value, use map for it

smay1613 (changes requested):

PR #190, opened 2020-12-16
ordered associative task, Siebova

annsv:


ordered associative task/student/include/impl.h
```diff
@@ -15,4 +15,14 @@
  *
  * @return 1-5
  */
-size_t convertMarkTo5PointSystem(size_t markIn100Point);
+size_t convertMarkTo5PointSystem(size_t markIn100Point)
+{
+    const std::map<int, int> marks {{29, 1}, {49, 2}, {69, 3}, {89, 4}};
+    int result = 5;
```
smay1613, line 21: please avoid the default value, use map

annsv: Fixed. Thank you!

ordered associative task/student/include/impl.h
```diff
@@ -17,12 +17,13 @@
  */
 size_t convertMarkTo5PointSystem(size_t markIn100Point)
 {
-    const std::map<int, int> marks {{29, 1}, {49, 2}, {69, 3}, {89, 4}};
-    int result = 5;
+    const std::map<int, int> marks {{29, 1}, {49, 2}, {69, 3}, {89, 4}, {100, 5}};
+    int result  = marks.upper_bound(markIn100Point) -> first;
```
smay1613: still not the easiest solution, remove this result at all and use upper_bound instead of lower_bound
re-consider intervals in this case

PR #201, opened 2020-12-17
Ordered Associative Task Zaiets

EugeneZaiets:


ordered associative task/student/include/impl.h
```diff
@@ -15,4 +14,16 @@
...
+    { 29 , 1 },
+    { 49 , 2 },
+    { 69 , 3 },
+    { 89 , 4 },
+};
+
+size_t convertMarkTo5PointSystem(size_t markIn100Point)
+{
+    auto it = association.lower_bound(markIn100Point);
+    return markIn100Point >= 90 ? 5 : it->second;
```
smay1613, line 28: dirty hack!
use map only without the additional magic numbers

smay1613 (changes requested):

PR #204, opened 2020-12-17
Adaptor task TDobrovolska

TetianaDobrovolska:


adaptors_task/student/src/impl.cpp
```diff
@@ -1,8 +1,38 @@
...
     {'[', ']'},
     {'{', '}'}
 };
+
+bool isValid(const std::string& source)
+{
+    std::stack<char> stack;
+    for (const char& c : source)
+    {
+        const auto it = std::find_if(brackets.cbegin(), brackets.cend(), [&](std::pair<char, char> const & pair)
```
smay1613, line 17: please name the predicate

adaptors_task/student/src/impl.cpp
```diff
@@ -1,8 +1,38 @@
...
+        {
+                return pair.first == c || pair.second == c;
+        });
+        if (it != brackets.end())
+        {
+            if (it->first == c)
```
smay1613, line 23: please name conditions

PR #207, opened 2020-12-17
unoredered associative container TDobrovolska

TetianaDobrovolska:


unordered_associative_task_1/student/src/impl.cpp
```diff
@@ -1 +1,30 @@
...
+    }
+
+    for (const char& c : first)
+    {
+        map_converter[c]++;
+    }
+
+    for (const char& c : second)
+    {
+        if (map_converter.end() == map_converter.find(c))
```
smay1613, line 20: why not plain map_converter[c]--?

unordered_associative_task_1/student/src/impl.cpp
```diff
@@ -1 +1,30 @@
...
+        {
+            return false;
+        }
+        if (0 == (--map_converter[c]))
+        {
+            map_converter.erase(c);
+        }
+    }
+    return map_converter.size() == 0;
```
smay1613, line 29: If you will go by approach like a stack (inc and decrement associated int) you can check if all "stacks" are 0

PR #208, opened 2020-12-18
adaptor task (amolodchenkova)

nussa909:


adaptors task/student/src/impl.cpp
```diff
@@ -1,8 +1,45 @@
...
     {'(', ')'},
     {'[', ']'},
     {'{', '}'}
 };
+
+bool isValid(const std::string& source)
+{
+    bool isValid = true;
+    std::stack<char> stack;
+    std::for_each(source.begin(),source.end(), [&stack,&isValid](const char& ch)
```
smay1613, line 16: don't use general std::for_each, always prefer more specific algorithm or use range for

adaptors task/student/src/impl.cpp
```diff
@@ -1,8 +1,45 @@
...
+    {
+       std::for_each(brackets.begin(),brackets.end(),[&ch,&stack,&isValid](const std::pair<char, char>& pair)
+       {
+          if(ch == pair.first)
+          {
+             stack.push(ch);
+          }
+
+          if(ch == pair.second)
```
smay1613, line 25: please name these conditions

adaptors task/student/src/impl.cpp
```diff
@@ -1,8 +1,45 @@
...
+          {
+              if ( stack.empty() )
+              {
+                  isValid = false;
+                  return;
+              }
+              else if( stack.top() != pair.first )
+              {
+                  isValid = false;
```
smay1613, line 34: why not return false?

adaptors task/student/src/impl.cpp
```diff
@@ -1,8 +1,45 @@
...
+          }
+
+          if(ch == pair.second)
+          {
+              if ( stack.empty() )
+              {
+                  isValid = false;
+                  return;
+              }
+              else if( stack.top() != pair.first )
```
smay1613, line 32: combine with previous and name the condition

smay1613 (changes requested):

PR #209, opened 2020-12-18
 ordered associative task (amolodchenkova)

nussa909:


ordered associative task/student/src/impl.cpp
```diff
@@ -1 +1,17 @@
 #include "impl.h"
+const std::map<size_t,size_t > marks{{0, 1},
+                                     {29, 1},
+                                     {30, 2},
+                                     {49, 2},
+                                     {50, 3},
```
smay1613, line 6: Hack!
Don't duplicate the values
Use intervals!

ordered associative task/student/src/impl.cpp
```diff
@@ -1 +1,17 @@
...
+                                     {69, 3},
+                                     {70, 4},
+                                     {89, 4},
+                                     {90, 5},
+                                     {100, 5}};
+
+size_t convertMarkTo5PointSystem(size_t markIn100Point)
+{
+    if(markIn100Point > 100) markIn100Point = 100;
```
smay1613, line 15: please complete task using map only

smay1613 (changes requested):

PR #214, opened 2020-12-20
Unordered associative task 2 Gershkul Aleksandr

DevASG:
There are some questions about this task description.
I didn't found a way how 
1. filtering_map object can be converted into "vector < string >";
2. std::tie in "bool operator <" can be applied for current task description;
3. it would be helpful to provide more details about the tasks.

### DIFF
--- a/unordered_associative_task_2/student/src/impl.cpp
+++ b/unordered_associative_task_2/student/src/impl.cpp
@@ -6,6 +6,17 @@
 #include <numeric>
 #include <vector>
 
+struct File {
+    std::string filename;
+    size_t size;
+    size_t hash;
+
+    bool operator<(const File& rhs)
+    {
+        return std::tie(size, hash) < std::tie(rhs.size, rhs.hash);
+    }
+};
+
 // use this for hashing
 auto xx_hash = [](const std::string& file) {
     std::ifstream stream { file };
@@ -35,11 +46,12 @@ auto filesize = [](const std::string& file) {
 template <class FilterCriteria>
 filtering_map filter(const std::vector<std::string>& files, FilterCriteria filter)
 {
-    filtering_map result;
+    filtering_map filteredFiles;
     for (auto const& file : files) {
         auto criteria = filter(file);
+        filteredFiles[criteria].push_back(file);
     }
-    return result;
+    return filteredFiles;
 }
 
 /** @note HELPER */
@@ -54,23 +66,33 @@ filtering_map groupDuplicates(const std::vector<std::string>& dataSource, Filter
 /** @todo With help of std filesystem, lists all regular files under specified directory recursively*/
 std::vector<std::string> listFiles(const std::string& directory)
 {
-    std::vector<std::string> files;
+    std::vector<std::string> fileNames;
     for (auto& p : fs::directory_iterator(directory)) {
         std::cout << p.path().filename().string() << '\n';
-        files.push_back(p.path());
+        fileNames.push_back(p.path());
     }
-    return files;
+    return fileNames;
 }
 
 /** @todo Implement function that will remove group in if it has only one or zero elements */
 void removeUniqueGroups(filtering_map& filteredData)
 {
+    for (auto& key_files : filteredData) {
+        if (key_files.second.size() == 1) {
+            filteredData.erase(key_files.first);
+        }
+    }
 }
 
 /** @todo Implement function that will transform map to a vector */
 std::vector<std::string> flattenGrouped(const filtering_map& grouped)
 {
     std::vector<std::string> result;
+    for (auto& key_files : grouped) {
+        for (auto& fileName : key_files.second) {
+            result.push_back(fileName);
+        }
+    }
     return result;
 }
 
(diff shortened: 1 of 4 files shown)


unordered_associative_task_2/student/src/impl.cpp
```diff
@@ -25,11 +25,42 @@ auto xx_hash = [] (const std::string &file)
     return xxh::xxhash<64>(buffer);
 };
 
+auto xx_file_size = [](const std::string& file) {
+    std::ifstream stream { file };
```
smay1613, line 29: Why do you need to open it?

unordered_associative_task_2/student/src/impl.cpp
```diff
@@ -41,9 +72,55 @@ filtering_map groupDuplicates (const std::vector<std::string>& dataSource, Filte
     return grouped;
 }
 
-std::vector<std::vector<std::string> > findDuplicates(const std::string &rootPath)
+/** @todo With help of std filesystem, lists all regular files under specified directory recursively*/
+std::vector<std::string> listFiles(const std::string& directory)
+{
+    std::vector<std::string> fileNames;
+    for (auto& p : fs::recursive_directory_iterator(directory)) {
```
smay1613, line 79: I would recommend fs::directory_options::skip_permission_denied option

smay1613, line 79: Simple and readable - nice :)
You can try to implement it in STL way by combining algorithms, iterators and functors.
Think what your loop does: copies value of iterator p to fileNames with condition (predicate).

unordered_associative_task_2/student/src/impl.cpp
```diff
@@ -41,9 +72,55 @@ filtering_map groupDuplicates (const std::vector<std::string>& dataSource, Filte
...
+}
+
+/** @todo Implement function that will transform map to a vector */
+std::vector<std::vector<std::string>> flattenGrouped(const filtering_map& grouped)
+{
+    std::vector<std::vector<std::string>> groups;
+    for (auto const& group : grouped) {
+        if (!group.second.empty()) {
+            std::vector<std::string> files(group.second.begin(), group.second.end());
+            groups.emplace_back(files);
```
smay1613, line 105: in this case, you should pass not the _files_, but _group.second.begin(), group.second.end()_
**You should pass initialization arguments for object to emplace function! Not the lvalue or rvalue!**

unordered_associative_task_2/student/src/impl.cpp
```diff
@@ -41,9 +72,55 @@ filtering_map groupDuplicates (const std::vector<std::string>& dataSource, Filte
...
+        }
+    }
+    return groups;
+}
+
+std::vector<std::vector<std::string>> findDuplicates(const std::string& rootPath)
 {
-    // filter by size
+    auto files = listFiles(rootPath);
```
smay1613, line 113: I would recommend const ref here

unordered_associative_task_2/student/src/impl.cpp
```diff
@@ -41,9 +72,55 @@ filtering_map groupDuplicates (const std::vector<std::string>& dataSource, Filte
...
+
+    // not filtered by size!
```
smay1613, line 115: Why?
It's a fine optimization

unordered_associative_task_2/student/src/impl.cpp
```diff
@@ -41,9 +72,55 @@ filtering_map groupDuplicates (const std::vector<std::string>& dataSource, Filte
...
+            ++it;
+    }
+}
+
+/** @todo Implement function that will transform map to a vector */
+std::vector<std::vector<std::string>> flattenGrouped(const filtering_map& grouped)
+{
+    std::vector<std::vector<std::string>> groups;
+    for (auto const& group : grouped) {
+        if (!group.second.empty()) {
```
smay1613, line 103: why do you need it?
iterators range will check the distance anyway

unordered_associative_task_2/student/src/impl.cpp
```diff
@@ -25,11 +25,42 @@ auto xx_hash = [] (const std::string &file)
...
+auto xx_file_size = [](const std::string& file) {
+    std::ifstream stream { file };
+    if (!stream) {
+        std::cerr << "File " << file << " can't be opened" << std::endl;
+        return 0ul;
+    }
+    return fs::file_size(file);
+};
+
+struct FileObject {
```
smay1613, line 37: Not used

unordered_associative_task_2/student/src/impl.cpp
```diff
@@ -25,11 +25,42 @@ auto xx_hash = [] (const std::string &file)
     return xxh::xxhash<64>(buffer);
 };
 
+auto xx_file_size = [](const std::string& file) {
```
smay1613, line 28: Not used

smay1613:
1. Implement size filtering and use flattening there :) _Hint: use insert inside_
2. Why do you think it should be applied here? Why do you need operator < here?

PR #215, opened 2020-12-21
resolving sequence task 3

collderart:
@smay1613 When I finally resolve it, I looked for resolving of other students and then thought that my solution is not what you have been expected...

But, I made it!)

### DIFF
--- a/sequence_task_3/student/src/impl.cpp
+++ b/sequence_task_3/student/src/impl.cpp
@@ -1,2 +1,17 @@
 #include "impl.h"
 
+std::string reverseWords(const std::string& input)
+{
+    if (input.empty()){
+        return input;
+    }
+    std::stringstream stream (input);
+    std::string streamOut;
+
+    for(std::string streamIn; stream>>streamIn; streamOut.insert(0, streamIn+' '));
+
+    streamOut.erase(streamOut.end() - 1);
+    if (*(streamOut.begin()) == ' ')
+        streamOut.erase(streamOut.begin());
+    return streamOut;
+}


sequence_task_3/student/src/impl.cpp
```diff
@@ -1,2 +1,17 @@
...
 
+std::string reverseWords(const std::string& input)
+{
+    if (input.empty()){
+        return input;
+    }
+    std::stringstream stream (input);
+    std::string streamOut;
+
+    for(std::string streamIn; stream>>streamIn; streamOut.insert(0, streamIn+' '));
```
smay1613, line 11: please use stream iterators

sequence_task_3/student/src/impl.cpp
```diff
@@ -1,2 +1,17 @@
...
+
+    streamOut.erase(streamOut.end() - 1);
+    if (*(streamOut.begin()) == ' ')
+        streamOut.erase(streamOut.begin());
```
smay1613, line 15: Not effective

sequence_task_3/student/src/impl.cpp
```diff
@@ -1,2 +1,17 @@
...
+    if (input.empty()){
+        return input;
+    }
+    std::stringstream stream (input);
+    std::string streamOut;
+
+    for(std::string streamIn; stream>>streamIn; streamOut.insert(0, streamIn+' '));
+
+    streamOut.erase(streamOut.end() - 1);
+    if (*(streamOut.begin()) == ' ')
```
smay1613, line 14: .front()

smay1613, line 14: you can use std::isspace

smay1613 (changes requested):

PR #216, opened 2020-12-26
pseudocontainers_Gershkul_Aleksandr

DevASG:


pseudocontainers_task/student/src/impl.cpp
```diff
@@ -1 +1,22 @@
 #include "impl.h"
+
+#include <bitset>
+#include <iostream>
+
+std::tuple<worker_id, worker_id, days> getMaxIntersection(const std::unordered_map<worker_id, calendar>& workerVacations)
+{
+    std::tuple<worker_id, worker_id, days> intersection(0, 0, 0);
+    for (auto iW = workerVacations.cbegin(); iW != workerVacations.cend(); ++iW) {
```
smay1613, line 9: use range for

pseudocontainers_task/student/src/impl.cpp
```diff
@@ -1 +1,22 @@
...
+        const std::bitset<daysInMonth> iVacation(iW->second, 0, daysInMonth, availableMarker, vacationMarker);
+
+        for (auto jW = std::next(iW); jW != workerVacations.cend(); ++jW) {
```
smay1613, line 12: range for

pseudocontainers_task/student/src/impl.cpp
```diff
@@ -1 +1,22 @@
...
+            const std::bitset<daysInMonth> jVacation(jW->second, 0, daysInMonth, availableMarker, vacationMarker);
+
+            auto ijIntersection = iVacation & jVacation;
+            if (std::get<2>(intersection) < ijIntersection.count()) {
+                intersection = std::tuple<worker_id, worker_id, days>(iW->first, jW->first, ijIntersection.count());
```
smay1613, line 17: use std::tie/std::make_tuple/std::forward_as_tuple

smay1613 (changes requested):

pseudocontainers_task/student/src/impl.cpp
DevASG: The main idea of using iterators in these for loops is to avoid overlapping of comparison and reduce O(n*n).
external for (i: 0..end() - 1) 
internal for (j: i + 1, end())
But using iterators leads to O(n*(n+1)/2) ~~ O(n*n)
So, "for range cycle" has almost the same complexity.

pseudocontainers_task/student/src/impl.cpp
DevASG: Done

pseudocontainers_task/student/src/impl.cpp
DevASG: Done

pseudocontainers_task/student/src/impl.cpp
smay1613: Please don't make premature optimizations that affects readability

smay1613: Upd: but, for internal loop It's okay here to do in such way. 
Sorry, my bad, internal loop can use iterators.

PR #221, opened 2021-02-15
[practice] iterators_task_1 implementation

opichuryn:


opichuryn:
better solution invented

PR #223, opened 2021-02-15
[ominenko]  functors_task_1

olemin7:
for lambda sameCountry
has return false if any country is Country::Unknown
UT updated

### DIFF
--- a/functors_task_1/student/include/impl.h
+++ b/functors_task_1/student/include/impl.h
@@ -1,13 +1,22 @@
 #pragma once
 #include "bar_serving.h"
 #include <functional>
-
+using namespace std;
 /**
  * @todo Implement functor-generator that will return next beer brand (cyclic)
  * @example HoeGaarden -> Corona -> Carlsberg -> Bud -> ZlataPraha -> Leffe -> HoeGaarden...
  */
 struct BeerOrganizer
 {
+    BeerBrand m_it=BeerBrand::None;
+public:
+    BeerBrand operator ()(){
+        m_it = (BeerBrand)(std::underlying_type<BeerBrand>::type(m_it) + 1);
+        if(BeerBrand::Max==m_it){
+            m_it=BeerBrand::HoeGaarden;
+        }
+        return m_it;
+    };
 };
 
 /**
@@ -16,16 +25,23 @@ struct BeerOrganizer
  *
  * @note Only Corona and HoeGaarden are expensive
  */
-bool isExpensiveBeer(/**???*/)
+bool isExpensiveBeer(const BeerBrand beerBrand)
 {
+    return (BeerBrand::Corona==beerBrand || BeerBrand::HoeGaarden==beerBrand);
 }
 
 /**
  * @todo Implement lambda beer country equality comparator
  * @return true if beer county is the same, false otherwise
  */
-auto sameCountry = [](/**???*/)
+auto sameCountry = [](const BeerBrand lhs,const BeerBrand rhs)
 {
+    const auto lCountry= getBeerCountry(lhs);
+    const auto rCountry= getBeerCountry(rhs);
+    if(Country::Unknown==lCountry){
+        return false;
+    }
+    return lCountry==rCountry;
 };
 
 struct MixingPolicy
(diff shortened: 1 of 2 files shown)


functors_task_1/student/include/impl.h
```diff
@@ -16,16 +25,23 @@ struct BeerOrganizer
...
 /**
  * @todo Implement lambda beer country equality comparator
  * @return true if beer county is the same, false otherwise
  */
-auto sameCountry = [](/**???*/)
+auto sameCountry = [](const BeerBrand lhs,const BeerBrand rhs)
 {
+    const auto lCountry= getBeerCountry(lhs);
+    const auto rCountry= getBeerCountry(rhs);
+    if(Country::Unknown==lCountry){
```
smay1613, line 41: Why do you think that false should be returned?
Also, why rhs county is not checked?

functors_task_1/student/include/impl.h
```diff
@@ -16,16 +25,23 @@ struct BeerOrganizer
...
+        return false;
+    }
+    return lCountry==rCountry;
```
smay1613, line 44: I suppose you can perform check inline here:
```suggestion
    return getBeerCountry(lhs) == getBeerCountry(rhs)
```

functors_task_1/student/include/impl.h
```diff
@@ -16,16 +25,23 @@ struct BeerOrganizer
  *
  * @note Only Corona and HoeGaarden are expensive
  */
-bool isExpensiveBeer(/**???*/)
+bool isExpensiveBeer(const BeerBrand beerBrand)
 {
+    return (BeerBrand::Corona==beerBrand || BeerBrand::HoeGaarden==beerBrand);
```
smay1613, line 30: _Up to your style, but you don't need brackets here_

functors_task_1/student/include/impl.h
```diff
@@ -40,9 +56,22 @@ struct MixingPolicy
      * Whiskey + SevenUp = SevenPlusSeven;
      * Others + Others = Oops;
      */
-    static Cocktail mix(/**???*/)
+    static Cocktail mix(const AlcoholDrink alcohol,const NonAlcoholDrink nonAlcohol)
     {
+        auto result=Cocktail::Oops;
+        if(AlcoholDrink::Gin==alcohol){
```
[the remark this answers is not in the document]
olemin7: switch is not good in scope code coverage

functors_task_1/student/include/impl.h
olemin7: no becouse of Country::Unknown

smay1613: Still don't get why do you think that Unknown is a special case

functors_task_1/student/include/impl.h
olemin7: descripion: return true if beer county is the same, false otherwise

Country::Unknown  - it's not a country, it union of some others country
that is why one country of one Country::Unknown != other Country::Unknown

checkin rhs is non nessessary becouse if lhs not Unknown and rhs Unknown false will be returned

functors_task_1/student/include/impl.h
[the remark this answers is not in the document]
smay1613: Nice concern, but could you please provide more info?

Actually, as for me, switching pattern here is needed because it can be easily replaced with a mapping pattern.
I mean a const static map, that represents recipes book. It is very convenient and extendable.

functors_task_1/student/include/impl.h
smay1613: "Unknown" represents a special invalid entry. This comparator should not care about invalid entries at all. It should just check for equality of properties, even if they are both invalid.

olemin7: there is many invalid variant can be
so there is has to be checking if country invalid
and if country are the same.

functors_task_1/student/include/impl.h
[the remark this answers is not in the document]
olemin7: pairs code static anaiser and unit test coverage
static - said that default must be included in switch
but if you use default - 100% brunch coverage is hard to reach
so you have to remove one opion or do some other trick... with comment like " analyzer petting"

and yes - swich can works faster and looks better than  if -else if

functors_task_1/student/include/impl.h
smay1613: it shouldn't be done by comparator

functors_task_1/student/include/impl.h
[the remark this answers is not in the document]
smay1613: Thanks for the details, but I think it should not be the case for our training.
Readability is the top#1 priority.

PR #225, opened 2021-02-15
[opichuryn] functors_task_1 implementation

opichuryn:


functors_task_1/student/include/impl.h
```diff
@@ -1,13 +1,33 @@
...
 
 /**
  * @todo Implement functor-generator that will return next beer brand (cyclic)
  * @example HoeGaarden -> Corona -> Carlsberg -> Bud -> ZlataPraha -> Leffe -> HoeGaarden...
  */
 struct BeerOrganizer
 {
+    BeerBrand operator()()
+    {
+        static std::vector<BeerBrand> brands = {
```
smay1613, line 14: 1. Please implement this task without containers usage
2. You don't need assignment
3. Why not const?
4. Why do you need vector, if the number of elements is known at compile time?
5. Why it is not auto-generated? What if there will be 50+ values?

functors_task_1/student/include/impl.h
```diff
@@ -40,9 +62,17 @@ struct MixingPolicy
      * Whiskey + SevenUp = SevenPlusSeven;
      * Others + Others = Oops;
      */
-    static Cocktail mix(/**???*/)
+    static Cocktail mix(AlcoholDrink alcohol, NonAlcoholDrink soft)
     {
+        if (alcohol == AlcoholDrink::Gin && soft == NonAlcoholDrink::LimeJuice) {
```
smay1613, line 67: Prefer switch

smay1613 (changes requested):

functors_task_1/student/include/impl.h
opichuryn: Done!

functors_task_1/student/include/impl.h
opichuryn: Done!

functors_task_1/student/include/impl.h
```diff
@@ -8,6 +8,26 @@
  */
 struct BeerOrganizer
 {
+    BeerBrand operator()()
+    {
+        constexpr size_t BEER_NUM = 6;
+        constexpr wall_shelf<BEER_NUM> brands {
```
smay1613, line 14: Still the problem of possible 50+ values and their initialization :)
I would prefer to iterate over enum, with range specification.
Also, what if the original BeerBrand enumeration will be changed?

functors_task_1/student/include/impl.h
```diff
@@ -16,16 +36,18 @@ struct BeerOrganizer
  *
  * @note Only Corona and HoeGaarden are expensive
  */
-bool isExpensiveBeer(/**???*/)
+bool isExpensiveBeer(BeerBrand brand)
```
smay1613, line 39: _const_

PR #226, opened 2021-02-16
functors_task_1_verna

VeeM:


functors_task_1/student/src/impl.cpp
```diff
@@ -1,2 +1,44 @@
 #include "impl.h"
 
+#include <algorithm>
+#include <forward_list>
+#include <map>
+#include <utility>
+
+BeerBrand BeerOrganizer::operator()()
+{
+    if (m_nextBeer <= static_cast<BaseType>(BeerBrand::None)
```
smay1613, line 10: I think it is better to specify range explicitly with const variables: "to" and "from"

functors_task_1/student/src/impl.cpp
```diff
@@ -1,2 +1,44 @@
...
+        m_nextBeer = static_cast<BaseType>(BeerBrand::None) + 1;
+    }
+
+    return static_cast<BeerBrand>(m_nextBeer++);
+}
+
+bool isExpensiveBeer(const BeerBrand brand)
+{
+    //TODO: move it somewhere
+    static std::forward_list<BeerBrand> expensiveBeer = {BeerBrand::Corona, BeerBrand::HoeGaarden};
```
smay1613, line 22: Wow, please implement this task without containers usage. Especially avoid lists (why - we'll talk about it at the second part of the training).

Don't overcomplicate here.

functors_task_1/student/src/impl.cpp
```diff
@@ -1,2 +1,44 @@
...
+    return std::find(std::begin(expensiveBeer), std::end(expensiveBeer), brand) != std::end(expensiveBeer);
+}
+
+Cocktail MixingPolicy::mix(const AlcoholDrink lhd, const NonAlcoholDrink rhd)
+{
+    //TODO: move it somewhere
+    static std::map<std::pair<AlcoholDrink, NonAlcoholDrink>, Cocktail> recipes = {
```
smay1613, line 29: 1. Yep, mapping is a good pattern and should be used here, but please, implement this task without containers usage. You can easily replace it with a switch.
2. Why not const?

smay1613 (changes requested):

functors_task_1/student/src/impl.cpp
smay1613: p.s also, you don't need ordered container here :)

functors_task_1/student/src/impl.cpp
VeeM: Yes, the forward list shouldn't be used here.

And the one more thing. How I should guess that it is forbidden to use containers? :)

functors_task_1/student/src/impl.cpp
VeeM: Should be const and unordered.
The same about containers as in the comment above nothing said about it in the task description. :)

functors_task_1/student/src/impl.cpp
VeeM: You're right! Thanks!

functors_task_1/student/include/impl.h
```diff
@@ -1,31 +1,41 @@
...
-{
-}
+bool isExpensiveBeer(const BeerBrand brand);
 
 /**
  * @todo Implement lambda beer country equality comparator
  * @return true if beer county is the same, false otherwise
  */
-auto sameCountry = [](/**???*/)
+auto sameCountry = [](BeerBrand lhd, BeerBrand rhd)
```
smay1613, line 36: _const here too :)_

functors_task_1/student/src/impl.cpp
```diff
@@ -1,2 +1,53 @@
...
+            result = Cocktail::Greyhount;
+        }
+        break;
+    case AlcoholDrink::Whiskey :
+        if (NonAlcoholDrink::SevenUp == rhd)
+        {
+            result = Cocktail::SevenPlusSeven;
+        }
+        break;
+    case AlcoholDrink::Beer :
```
smay1613, line 46: _You can skip that_

functors_task_1/student/include/impl.h
VeeM: О_о
Thanks :)

functors_task_1/student/src/impl.cpp
VeeM: But in this case, we will get a warning(something like "not all enum entries was handled") ...

smay1613: Yep, and it is okay here, as for me

PR #227, opened 2021-02-16
[Fedorovskyi] Implement first task

tra1ning5:


functors_task_1/student/include/impl.h
```diff
@@ -1,31 +1,57 @@
...
+        if(currentIndex == static_cast<size_t>(BeerBrand::Max))
+        {
+            currentIndex = 1;
+        }
+
+        return static_cast<BeerBrand>(currentIndex++);
+    }
 
+private:
+    size_t currentIndex {1};
```
smay1613, line 23: Please avoid magic numbers

functors_task_1/student/include/impl.h
```diff
@@ -1,31 +1,57 @@
...
 /**
  * @todo Implement functor-generator that will return next beer brand (cyclic)
  * @example HoeGaarden -> Corona -> Carlsberg -> Bud -> ZlataPraha -> Leffe -> HoeGaarden...
  */
 struct BeerOrganizer
 {
-};
+    BeerBrand operator()()
+    {
+        if(currentIndex == static_cast<size_t>(BeerBrand::Max))
```
smay1613, line 14: Specify the possible range with "to" and "from" const variables.

functors_task_1/student/include/impl.h
```diff
@@ -1,31 +1,57 @@
...
+        {
+            currentIndex = 1;
```
smay1613, line 16: why 1?

functors_task_1/student/include/impl.h
```diff
@@ -1,31 +1,57 @@
...
  */
-bool isExpensiveBeer(/**???*/)
+bool isExpensiveBeer(BeerBrand beer)
 {
+    switch(beer)
+    {
+    case BeerBrand::HoeGaarden:
+    case BeerBrand::Corona:
+        return true;
+    case BeerBrand::None:
```
smay1613, line 38: why should you manually specify all default cases?

functors_task_1/student/include/impl.h
```diff
@@ -40,9 +66,31 @@ struct MixingPolicy
      * Whiskey + SevenUp = SevenPlusSeven;
      * Others + Others = Oops;
      */
-    static Cocktail mix(/**???*/)
+    static Cocktail mix(AlcoholDrink alchohol, NonAlcoholDrink nonAlcohol)
+    {
+        std::unordered_map<AlcoholDrink, std::unordered_map<NonAlcoholDrink, Cocktail>> cocktailes =
```
smay1613, line 71: 1. Yep, mapping is a good pattern and should be used here, but please, implement this task without containers usage. You can easily replace it with a switch.
2. Why not const?
3. Why do you recreate this map on every function call?
4. You can remove "="

functors_task_1/student/include/impl.h
```diff
@@ -40,9 +66,31 @@ struct MixingPolicy
...
+        {
+            { AlcoholDrink::Gin,     { {NonAlcoholDrink::LimeJuice, Cocktail::Gimlet},
+                                       {NonAlcoholDrink::GrapefruitJuice, Cocktail::Greyhount} } },
+
+            { AlcoholDrink::Whiskey, { {NonAlcoholDrink::SevenUp, Cocktail::SevenPlusSeven} } }
+        };
+
+        if( isFound(cocktailes, alchohol) &&
```
smay1613, line 79: Looks like you set wrong type for a receipt book: the key is the pair of alcohol and non-alcohol drink, and result is the value.

functors_task_1/student/include/impl.h
```diff
@@ -40,9 +66,31 @@ struct MixingPolicy
...
+        {
+            return cocktailes[alchohol][nonAlcohol];
+        }
+
+        return Cocktail::Oops;
+    }
+
+private:
+    template<class Map, class Key>
+    static bool isFound(const Map& map, const Key& key)
```
smay1613, line 90: I suppose it should be named as "contains"

functors_task_1/student/include/impl.h
```diff
@@ -40,9 +66,31 @@ struct MixingPolicy
...
+        std::unordered_map<AlcoholDrink, std::unordered_map<NonAlcoholDrink, Cocktail>> cocktailes =
+        {
+            { AlcoholDrink::Gin,     { {NonAlcoholDrink::LimeJuice, Cocktail::Gimlet},
+                                       {NonAlcoholDrink::GrapefruitJuice, Cocktail::Greyhount} } },
+
+            { AlcoholDrink::Whiskey, { {NonAlcoholDrink::SevenUp, Cocktail::SevenPlusSeven} } }
+        };
+
+        if( isFound(cocktailes, alchohol) &&
+                isFound(cocktailes[alchohol], nonAlcohol) )
```
smay1613, line 80: if you already checked that the map contains value, why do you need to use "[]" operator?

smay1613 (changes requested):

functors_task_1/student/include/impl.h
```diff
@@ -40,9 +54,28 @@ struct MixingPolicy
...
+        {
+        case AlcoholDrink::Gin:
+            switch(nonAlcohol)
+            {
+            case NonAlcoholDrink::LimeJuice:
+                return Cocktail::Gimlet;
+            case NonAlcoholDrink::GrapefruitJuice:
+                return Cocktail::Greyhount;
+            default:
+                return Cocktail::Oops;
```
smay1613, line 69: I'd prefer moving it to the end of function and simply break here

PR #231, opened 2021-02-16
functors_task_1_Serhii_Starodumov

SStarodumov:


functors_task_1/student/include/impl.h
```diff
@@ -8,6 +8,11 @@
  */
 struct BeerOrganizer
 {
+    BeerOrganizer();
+    size_t m_number;
```
smay1613, line 12: _please use more descriptive name_

functors_task_1/student/include/impl.h
```diff
@@ -8,6 +8,11 @@
  */
 struct BeerOrganizer
 {
+    BeerOrganizer();
```
smay1613, line 11: Use in-place initialization:
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c45-dont-define-a-default-constructor-that-only-initializes-data-members-use-in-class-member-initializers-instead

functors_task_1/student/src/impl.cpp
```diff
@@ -1,2 +1,37 @@
...
+    }
+    return static_cast<BeerBrand>(m_number++);
+}
+
+BeerOrganizer::BeerOrganizer():m_number(m_min+1) {
+    static_assert(m_max-m_min>1, "You have only None and Max in enum!");
+}
+
+bool isExpensiveBeer(const BeerBrand brand) {
+    constexpr static std::array<BeerBrand, 2> expensiveBeers =
```
smay1613, line 17: I suppose that is overengineering, please simplify with plain logical expression

functors_task_1/student/src/impl.cpp
```diff
@@ -1,2 +1,37 @@
...
+            {BeerBrand::Corona, BeerBrand::HoeGaarden};
+    return std::find(expensiveBeers.cbegin(), expensiveBeers.cend(), brand)!=expensiveBeers.cend();
+}
+
+Cocktail MixingPolicy::mix(const AlcoholDrink alcoholDrink, const NonAlcoholDrink nonAlcoholDrink) {
+    using Ingredients = std::pair<AlcoholDrink, NonAlcoholDrink>;
+    using Recipe = std::pair<Ingredients , Cocktail>;
+    constexpr static std::array<Recipe , 3> recipes = {
```
smay1613, line 25: In such cases, unordered map should be preferred.

But nice try :)

smay1613 (changes requested):

functors_task_1/student/src/impl.cpp
```diff
@@ -1,2 +1,37 @@
...
+}
+
+bool isExpensiveBeer(const BeerBrand brand) {
+    constexpr static std::array<BeerBrand, 2> expensiveBeers =
+            {BeerBrand::Corona, BeerBrand::HoeGaarden};
+    return std::find(expensiveBeers.cbegin(), expensiveBeers.cend(), brand)!=expensiveBeers.cend();
+}
+
+Cocktail MixingPolicy::mix(const AlcoholDrink alcoholDrink, const NonAlcoholDrink nonAlcoholDrink) {
+    using Ingredients = std::pair<AlcoholDrink, NonAlcoholDrink>;
```
smay1613, line 23: +1 for aliases, they make code more readable

functors_task_1/student/src/impl.cpp
SStarodumov: according to cppref (https://en.cppreference.com/w/cpp/utility/hash) std::pair has no std::hash

smay1613: So you should provide your own implementation, it is pretty simple

PR #235, opened 2021-02-17
functors task (1) - Pysarenko

Elzianor:


functors_task_1/student/include/impl.h
```diff
@@ -1,13 +1,21 @@
...
 /**
  * @todo Implement functor-generator that will return next beer brand (cyclic)
  * @example HoeGaarden -> Corona -> Carlsberg -> Bud -> ZlataPraha -> Leffe -> HoeGaarden...
  */
 struct BeerOrganizer
 {
+    BeerBrand operator()();
+
+private:
+    size_t currentIndex{1};
```
smay1613, line 18: avoid magic numbers

functors_task_1/student/include/impl.h
```diff
@@ -16,16 +24,15 @@ struct BeerOrganizer
...
-{
-}
+bool isExpensiveBeer(const BeerBrand& brand);
 
 /**
  * @todo Implement lambda beer country equality comparator
  * @return true if beer county is the same, false otherwise
  */
-auto sameCountry = [](/**???*/)
+auto sameCountry = [](const BeerBrand& lhs, const BeerBrand& rhs)
```
smay1613, line 33: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f16-for-in-parameters-pass-cheaply-copied-types-by-value-and-others-by-reference-to-const

functors_task_1/student/src/impl.cpp
```diff
@@ -1,2 +1,29 @@
 #include "impl.h"
 
+BeerBrand BeerOrganizer::operator()() {
+    if (static_cast<size_t>(BeerBrand::Max) == currentIndex)
+        currentIndex = 1;
```
smay1613, line 5: avoid magic numbers, specify "from" and "to" const variables

functors_task_1/student/src/impl.cpp
```diff
@@ -1,2 +1,29 @@
...
+
+    return static_cast<BeerBrand>(currentIndex++);
+}
+
+bool isExpensiveBeer(const BeerBrand& brand)
+{
+    std::vector<BeerBrand> expensive{BeerBrand::HoeGaarden, BeerBrand::Corona};
```
smay1613, line 12: 1. Why do you need dynamic container?
2. Why not the plain logical expression?

functors_task_1/student/src/impl.cpp
```diff
@@ -1,2 +1,29 @@
...
+
+    return std::find(expensive.begin(), expensive.end(), brand) != std::end(expensive);
+}
+
+Cocktail MixingPolicy::mix(AlcoholDrink alcohol, NonAlcoholDrink nonAlcohol)
+{
+    using Recipe = std::tuple<AlcoholDrink, NonAlcoholDrink>;
+    using Ingredients = std::pair<AlcoholDrink, NonAlcoholDrink>;
+
+    std::map<Recipe, Cocktail> cookbook{{Ingredients(AlcoholDrink::Gin, NonAlcoholDrink::LimeJuice), Cocktail::Gimlet},
```
smay1613, line 22: 1. In such cases, unordered_map should be used, as you don't need the order
2. Why you need to recreate this object every time function is called?
3. Mark it as const

functors_task_1/student/src/impl.cpp
```diff
@@ -1,2 +1,29 @@
...
+bool isExpensiveBeer(const BeerBrand& brand)
+{
+    std::vector<BeerBrand> expensive{BeerBrand::HoeGaarden, BeerBrand::Corona};
+
+    return std::find(expensive.begin(), expensive.end(), brand) != std::end(expensive);
+}
+
+Cocktail MixingPolicy::mix(AlcoholDrink alcohol, NonAlcoholDrink nonAlcohol)
+{
+    using Recipe = std::tuple<AlcoholDrink, NonAlcoholDrink>;
```
smay1613, line 19: why tuple and not the pair?

smay1613 (changes requested):

functors_task_1/student/include/impl.h
Elzianor: Changed to use enum value

functors_task_1/student/include/impl.h
Elzianor: Const and & removed.

functors_task_1/student/src/impl.cpp
Elzianor: Changed to use enum values

functors_task_1/student/src/impl.cpp
Elzianor: The purpose was to reuse existing find algorithm, but in this case it seems to be an overhead. Changed to switch - case.

functors_task_1/student/src/impl.cpp
Elzianor: Changed to static const unordered_map.

functors_task_1/student/src/impl.cpp
Elzianor: Tuple was removed, now only pair is used.

functors_task_1/student/src/impl.cpp
```diff
@@ -2,28 +2,44 @@
...
+bool isExpensiveBeer(BeerBrand brand)
 {
-    std::vector<BeerBrand> expensive{BeerBrand::HoeGaarden, BeerBrand::Corona};
-
-    return std::find(expensive.begin(), expensive.end(), brand) != std::end(expensive);
+    switch(brand)
+    {
+    case BeerBrand::HoeGaarden:
+    case BeerBrand::Corona:
+        return true;
```
smay1613, line 16: avoid explicit true/false in your code

functors_task_1/student/src/impl.cpp
```diff
@@ -2,28 +2,44 @@
...
+    using Recipe = std::pair<AlcoholDrink, NonAlcoholDrink>;
+
+    struct hashPair
+    {
+        size_t operator()(const Recipe& recipe) const
+        {
+            size_t first = static_cast<size_t>(recipe.first);
+            size_t second = static_cast<size_t>(recipe.second);
+
+            size_t cantorPairing = (first + second) * (first + second + 1) / 2 + second;
```
smay1613, line 33: prefer std::hash

functors_task_1/student/src/impl.cpp
```diff
@@ -2,28 +2,44 @@
...
+    }
 }
 
 Cocktail MixingPolicy::mix(AlcoholDrink alcohol, NonAlcoholDrink nonAlcohol)
 {
-    using Recipe = std::tuple<AlcoholDrink, NonAlcoholDrink>;
-    using Ingredients = std::pair<AlcoholDrink, NonAlcoholDrink>;
+    using Recipe = std::pair<AlcoholDrink, NonAlcoholDrink>;
+
+    struct hashPair
```
smay1613, line 26: extend std namespace with hash specialization

functors_task_1/student/include/impl.h
```diff
@@ -15,7 +15,7 @@ struct BeerOrganizer
     BeerBrand operator()();
 
 private:
-    size_t currentIndex{1};
+    size_t currentIndex{static_cast<size_t>(BeerBrand::HoeGaarden)};
```
smay1613, line 18: please specify range "from" and "to"

functors_task_1/student/src/impl.cpp
```diff
@@ -2,28 +2,44 @@
 
 BeerBrand BeerOrganizer::operator()() {
     if (static_cast<size_t>(BeerBrand::Max) == currentIndex)
```
smay1613, line 4: extract value to a named const variable

smay1613 (changes requested):

functors_task_1/student/src/impl.cpp
Elzianor: Could you please explain why?

functors_task_1/student/src/impl.cpp
Elzianor: Xor is giving the same result in cases of (a xor b) and (b xor a).
So for example pairs (OrangeJuice = 1, Wine = 2) and (GrapefruitJuice = 2, Whiskey = 1) will result in the same hash value which is wrong. Cantor pairing approach on the other hand handles such cases correctly.

functors_task_1/student/src/impl.cpp
smay1613: Doing so avoids verbosity and eliminates some opportunities for mistakes. Helps make style consistent and conventional.

functors_task_1/student/src/impl.cpp
smay1613: no problem, but make it more explicit - extract to some function, I've suggested xor for simplicity

PR #237, opened 2021-02-18
iterators_task_1_Serhii_Starodumov

SStarodumov:


iterators_task_1/student/include/impl.h
```diff
@@ -11,3 +12,17 @@
...
+public:
+    const std::string& getToken() const;
+    std::string& getToken();
+    bool isEmpty() const;
+private:
+    std::string m_data;
+};
+
+std::istream& operator>>(std::istream& stream, TokenString& token){
+    std::getline(stream, token.getToken(), ' ');
```
smay1613, line 26: space is a default separator for strings in stream, so I think you can remove this class at all and use default std::string

iterators_task_1/student/src/impl.cpp
```diff
@@ -1,2 +1,39 @@
 #include "impl.h"
+#include <vector>
+#include <algorithm>
 
+std::string reverseWords(const std::string& input){
+    std::stringstream sstream(input);
+    std::vector<TokenString> tokens {std::istream_iterator<TokenString>{sstream}, std::istream_iterator<TokenString>{}};
+
+
+    tokens.erase(std::remove_if(tokens.begin(), tokens.end(), [](TokenString token){
```
smay1613, line 10: 1. Why do you copy TokenString?
2. String stream iterator should automatically skip empty tokens

iterators_task_1/student/src/impl.cpp
```diff
@@ -1,2 +1,39 @@
...
+        return token.isEmpty();
+    }), tokens.end());
+
+    std::vector<std::string> reversedTokens;
+    std::transform(tokens.rbegin(), tokens.rend(), std::back_inserter(reversedTokens), [](const TokenString& token){
+        return token.getToken();
+    });
+
+    sstream = std::stringstream();
```
smay1613, line 19: please don't:
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es26-dont-use-a-variable-for-two-unrelated-purposes

smay1613 (changes requested):

PR #238, opened 2021-02-18
iterators_task_1_verna

VeeM:


iterators_task_1/student/src/impl.cpp
```diff
@@ -1,2 +1,62 @@
...
+std::string reverseWords(const std::string& input)
+{
+    if (input.empty())
+    {
+        return {};
+    }
+
+    std::istringstream inputStream {input};
+    std::ostringstream outputStream;
+    std::list<std::string> container = {};
```
VeeM, line 18: @smay1613 , what about the list for this case (to avoid reallocation in the vector for long input string)?

iterators_task_1/student/src/impl.cpp
```diff
@@ -1,2 +1,62 @@
...
+              std::front_inserter(container));
+
+    if (container.empty())
+    {
+        return {};
+    }
+
+    auto begin = std::begin(container);
+    auto end = std::end(container);
+    std::copy(begin, std::prev(end), std::ostream_iterator<std::string>(outputStream, " "));
```
VeeM, line 31: @smay1613 , maybe a stupid question, but why std::move(algorithm) does not work in this case (does not move container content)?

iterators_task_1/student/src/impl.cpp
```diff
@@ -1,2 +1,62 @@
...
+//        return {};
+//    }
+
+//    auto rbegin = container.rbegin();
+//    auto rend = container.rend();
+//    std::copy(rbegin, std::prev(rend), std::ostream_iterator<std::string>(outputStream, " "));
+//    std::copy(std::prev(rend), rend, std::ostream_iterator<std::string>(outputStream));
+
+//    return outputStream.str();
+//}
```
VeeM, line 62: @smay1613 , the same but with vector :)

iterators_task_1/student/src/impl.cpp
```diff
@@ -1,2 +1,59 @@
...
+
+    if (container.empty())
+    {
+        return {};
+    }
+
+    auto rbegin = container.rbegin();
+    auto rend = container.rend();
+    std::copy(rbegin, std::prev(rend), std::ostream_iterator<std::string>(outputStream, " "));
+    std::copy(std::prev(rend), rend, std::ostream_iterator<std::string>(outputStream));
```
VeeM, line 29: @smay1613
Would be better just to pop the last space from the result?

iterators_task_1/student/src/impl.cpp
```diff
@@ -1,2 +1,59 @@
...
+std::string reverseWords(const std::string& input)
+{
+    if (input.empty())
+    {
+        return {};
+    }
+
+    std::istringstream inputStream {input};
+    std::ostringstream outputStream;
+    std::list<std::string> container { std::istream_iterator<std::string>(inputStream),
```
smay1613, line 18: please avoid std::list usage

iterators_task_1/student/src/impl.cpp
smay1613: operator <<

iterators_task_1/student/src/impl.cpp
```diff
@@ -1,2 +1,59 @@
...
+
+std::string reverseWords(const std::string& input)
+{
+    if (input.empty())
+    {
+        return {};
+    }
+
+    std::istringstream inputStream {input};
+    std::ostringstream outputStream;
```
smay1613, line 17: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es22-dont-declare-a-variable-until-you-have-a-value-to-initialize-it-with

smay1613 (changes requested):

PR #239, opened 2021-02-18
[Maksym Rassokha] Iterator_task_1

mrassokh:


iterators_task_1/student/include/impl.h
```diff
@@ -10,4 +12,32 @@
  *
  * @warning YOU MUST USE STRING STREAMS AND STREAM ITERATORS!
  */
-std::string reverseWords(const std::string& input);
+struct line : public std::string {};
+std::istream & operator >> (std::istream& stream, line& token)
+{
+    std::getline(stream,token,' ');
```
smay1613, line 18: 1. That is not line reading
2. By default, stream delimiter is a space. So you don't need that at all

iterators_task_1/student/include/impl.h
```diff
@@ -10,4 +12,32 @@
...
+    return stream;
+}
+
+auto isNonZeroString = [](std::string const & str) {
+    return str.size() != 0;
```
smay1613, line 23: .empty()

iterators_task_1/student/include/impl.h
```diff
@@ -10,4 +12,32 @@
...
+std::string reverseWords(const std::string& input)
+{
+    std::stringstream streamInput(input);
+    std::vector<std::string> tokenized(std::istream_iterator<line>{streamInput},
+                                       std::istream_iterator<line>{});
+
+    std::stringstream streamOutput{};
+    std::copy_if(tokenized.rbegin(),tokenized.rend(),
+                 std::ostream_iterator<std::string>{streamOutput," "},
+                 isNonZeroString);
```
smay1613, line 35: string stream skips spaces automatically

smay1613 (changes requested):

PR #241, opened 2021-02-18
[ominenko] iterators_task_1

olemin7:


iterators_task_1/student/src/impl.cpp
```diff
@@ -1,2 +1,21 @@
 #include "impl.h"
+#include "vector"
 
+std::string reverseWords(const std::string& input){
+    std::istringstream stream {input};
+    std::vector<std::string> words{std::istream_iterator<std::string> {stream},
+                                   std::istream_iterator<std::string> {}};
+    std::ostringstream out;
+    bool fdelim=false;
+    for(auto it=words.rbegin();it!=words.rend();it++){
```
smay1613, line 10: 1. Avoid loops
2. Use output stream iterators

PR #242, opened 2021-02-18
iterators_task_1_kostiuchenko

kosnko:


iterators_task_1/student/src/impl.cpp
```diff
@@ -1,2 +1,17 @@
 #include "impl.h"
+#include <list>
 
+std::string reverseWords(const std::string& input)
+{
+    std::stringstream inputStream{input};
+    std::stringstream outputStream;
+    std::list<std::string> inputStrList;
```
smay1613, line 8: please avoid std::list!
use deque instead

kosnko (commented):
changed list to deque

PR #247, opened 2021-02-19
[Maksym Rassokha] Functor_Task_2

mrassokh:


functors_task_2/student/include/impl.h
```diff
@@ -62,10 +64,15 @@ class DataBrowser
...
-        // find reader
-        // check for errors
-        // call functor
-        return false;
+        auto readerIterator = m_dataReaders.find(userId);
+
+        if (readerIterator == m_dataReaders.end())
+            return false;
+
+        if (readerIterator->second == nullptr)
```
smay1613, line 72: depends on style, but you can use implicit conversion to bool, it will look much simpler

PR #252, opened 2021-02-21
functors_task_2_verna

VeeM:


functors_task_2/student/include/impl.h
```diff
@@ -76,8 +81,19 @@ class DataBrowser
     template<class Functional, typename Output>
     bool invokeDataRequest(Functional method, const std::unique_ptr<IDataSelector>& selector, Output& result) const
     {
-        // adapt function
-        // call selector member
+        if (selector)
+        {
+            return std::ref(method)(selector.get(), result);
```
smay1613, line 86: please leave the "normal" execution flow till the end, and check the preconditions
so, swap line 86 with line 89

PR #262, opened 2021-02-21
Iterators_task_2_Serhii_Starodumov

SStarodumov:


iterators_task_2/student/include/impl.h
```diff
@@ -2,14 +2,45 @@
 #include <iterator>
 
 template <class Container>
-struct accumulator
+class accumulator : public std::iterator<std::output_iterator_tag, typename Container::value_type>
```
smay1613, line 5: does the value type has sense?

PR #266, opened 2021-02-22
[Fedorovskyi] Iterator task 2

tra1ning5:


iterators_task_2/student/include/impl.h
```diff
@@ -1,15 +1,47 @@
...
+
+    accumulator(Container& conatiner) :
+        m_container{conatiner}
+    {}
+
+
+    template<class OtherConatianer>
+    accumulator& operator =(OtherConatianer&& rhs)
+    {
+        std::move(std::make_move_iterator(rhs.begin()),
```
smay1613, line 23: you don't need move iterators with std::move algorithm

PR #267, opened 2021-02-22
algorithms_task_1_ominenko

olemin7:


algorithms_task_1/student/include/impl.h
```diff
@@ -11,4 +11,10 @@
   * @return consecutive range
 */
 template<class FI, class Comparator = std::equal_to<typename std::iterator_traits<FI>::value_type>>
-std::pair<FI, FI> consecutive_group(FI start, FI end, Comparator comp = {});
+std::pair<FI, FI> consecutive_group(FI start, FI end, Comparator comp = {}){
+    auto beg=std::adjacent_find(start,end,comp);
+    auto last=std::find_if_not(beg,end,[=](typename std::iterator_traits<FI>::value_type val){
```
smay1613, line 16: don't copy, use const ref

PR #268, opened 2021-02-22
[opichuryn] algorithms_task_2

opichuryn:


algorithms_task_2/student/src/impl.cpp
```diff
@@ -1 +1,15 @@
...
+{
+    if (prefix.empty()) {
+        return true;
+    }
+
+    if (prefix.size() > source.size()) {
+        return false;
+    }
+
+    return std::equal(source.begin(), source.begin() + prefix.size(), prefix.begin());
```
smay1613, line 14: _what if source will be empty?
operator + will result UB for end iterator

looks like it is safe here, but I'd make it safer without any shifts that possibly may result an error_

opichuryn: Done!

PR #270, opened 2021-02-22
[Fedorovskyi] algorithm_task_2

tra1ning5:


algorithms_task_2/student/src/impl.cpp
```diff
@@ -1 +1,6 @@
 #include "impl.h"
+
+bool startsWith(const std::string &source, const std::string &prefix)
+{
+    return source.find(prefix) != std::string::npos;
```
smay1613, line 5: has overhead, prefer algorithms version

p.s: think why it would have an overhead

tra1ning5: Overhead, because string::find has O(prefix.size() * source.size()) complexity, but std::equal has at most last1 - first1 applications of the predicate for C++11

algorithms_task_2/student/src/impl.cpp
```diff
@@ -1 +1,10 @@
 #include "impl.h"
+#include <algorithm>
+
+bool startsWith(const std::string &source, const std::string &prefix)
+{
+    if(prefix.size() > source.size())
+        return false;
+
+    return std::equal(begin(prefix), end(prefix), begin(source));
```
smay1613, line 9: Did you rely on the ADL accidentally?

PR #274, opened 2021-02-22
algorithms_task_2 Chiruk

ChNokia:


algorithms_task_2/student/include/impl.h
```diff
@@ -9,4 +10,13 @@
  *
  * @return true if source starts from a prefix
  */
-bool startsWith(const std::string& source, const std::string& prefix);
+bool startsWith(const std::string& source, const std::string& prefix) {
+    if (prefix.empty()) {
+        return prefix.empty() || source.empty();
```
smay1613, line 15: why do you need to check both? it will be evaluated to true

algorithms_task_2/student/include/impl.h
```diff
@@ -9,4 +10,13 @@
...
+    }
+    if (prefix.size() > source.length()) {
```
smay1613, line 17: please prefer one style: length() or size()

algorithms_task_2/student/include/impl.h
```diff
@@ -9,4 +10,13 @@
...
+        return false;
+    }
+
+    return std::equal(source.cbegin(), source.cbegin() + prefix.length(), prefix.cbegin());
```
smay1613, line 21: not the case here, but please make it safer, without any iterator shifts

algorithms_task_2/student/include/impl.h
ChNokia: my mistake

algorithms_task_2/student/include/impl.h
ChNokia: done

algorithms_task_2/student/include/impl.h
ChNokia: done

PR #275, opened 2021-02-22
[Maksym Rassokha] Algorithm_Task_2

mrassokh:


algorithms_task_2/student/include/impl.h
```diff
@@ -9,4 +10,18 @@
...
+{
+// possible solution:
+//    int i = 0;
+//    while (i < prefix.size() && prefix[i] == source[i]){++i;}
+//    return i == prefix.length();
+
+  if (prefix.size() > source.size())
+      return false;
+
+  auto mismatch = std::mismatch(source.begin(), source.begin() + prefix.size(),
```
smay1613, line 23: nice idea, but we don't need iterators here, prefer equal :)

algorithms_task_2/student/include/impl.h
```diff
@@ -9,4 +10,16 @@
...
+{
+// possible solution:
+//    int i = 0;
+//    while (i < prefix.size() && prefix[i] == source[i]){++i;}
+//    return i == prefix.length();
+
+  if (prefix.size() > source.size())
+      return false;
+
+  return std::equal(source.begin(), std::next(source.begin(),prefix.size()),
```
smay1613, line 23: I'd make it safer, without any iterator shifts

PR #276, opened 2021-02-22
iterators_task_2_kostiuchenko

kosnko:


iterators_task_2/student/include/impl.h
```diff
@@ -1,15 +1,49 @@
...
+    }
+
+    accumulator& operator=(Container&& conRR) {
+        container->insert(container->end(),
+                          std::make_move_iterator(conRR.begin()),
+                          std::make_move_iterator(conRR.end()));
+        return *this;
+    }
+
+    accumulator& operator=(const std::deque<typename Container::value_type>& decqu) {
```
smay1613, line 32: why do you need it?

iterators_task_2/student/include/impl.h
```diff
@@ -1,15 +1,49 @@
...
+        container->insert(container->end(), decqu.begin(), decqu.end());
+        return *this;
+    }
+
+    accumulator& operator=(const typename Container::value_type& value) {
```
smay1613, line 37: This iterator does not support value insertion

iterators_task_2/student/include/impl.h
```diff
@@ -1,15 +1,49 @@
...
 {
     /** @todo Iterator traits*/
     /** @todo Constructor with container*/
     /** @todo lvalue writing operator*/
     /** @todo rvalue writing operator*/
 
     /** @todo Iterator operations */
+
+protected:
+    Container* container;
```
smay1613, line 16: What is the reason of usage raw pointer?

iterators_task_2/student/include/impl.h
```diff
@@ -1,15 +1,49 @@
...
+public:
+    explicit accumulator(Container& con): container(&con) {};
+
+    accumulator& operator=(Container& conLR) {
```
smay1613, line 20: You already have a template version

kosnko (commented):
If I would like to specify the same value_type for another container. Is it correct?

iterators_task_2/student/include/impl.h
kosnko: I would like to specify the same value_type for another container but assertion failed

iterators_task_2/student/include/impl.h
kosnko: removed

iterators_task_2/student/include/impl.h
```diff
@@ -13,35 +12,31 @@ struct accumulator : public std::iterator<std::output_iterator_tag, void, void,
...
-        return *this;
-    }
-
-    accumulator& operator=(Container&& conRR) {
-        container->insert(container->end(),
-                          std::make_move_iterator(conRR.begin()),
-                          std::make_move_iterator(conRR.end()));
+    explicit accumulator(Container& con): container(con) {};
+
+    // ??? If I would like to specify the same value_type for another container. Is it correct?
```
smay1613, line 19: I think it is not a good practice, because of implicitly converted types. This will reduce possible applications of our template

PR #285, opened 2021-02-23
algorithms_task_2_kostiuchenko

kosnko:


algorithms_task_2/student/include/impl.h
```diff
@@ -9,4 +9,12 @@
  *
  * @return true if source starts from a prefix
  */
-bool startsWith(const std::string& source, const std::string& prefix);
+bool startsWith(const std::string& source, const std::string& prefix)
+{
+    if (prefix.size() > source.size()) return false;
```
smay1613, line 14: don't write expressions on the same line

algorithms_task_2/student/include/impl.h
```diff
@@ -9,4 +9,12 @@
...
+    return prefix.cend() == std::mismatch(prefix.cbegin(), prefix.cend(), source.cbegin()).first;
+
+    //alternative
+//    if (prefix.size() > source.size()) return false;
+//    return std::equal(prefix.cbegin(), prefix.cend(), source.cbegin());
```
smay1613, line 19: This is better alternative for this case, because it does not require checking the iterator

PR #290, opened 2021-02-24
functors_task_1

viktorbezkrovniy:


functors_task_1/student/include/impl.h
```diff
@@ -8,6 +8,18 @@
  */
 struct BeerOrganizer
 {
+    size_t m_start = static_cast<size_t> (BeerBrand::HoeGaarden);
```
smay1613, line 11: 1. Const
2. It is better to bind these values to a "special" entries like "Max" or "None"

functors_task_1/student/include/impl.h
```diff
@@ -16,16 +28,18 @@ struct BeerOrganizer
...
 {
+    return (BeerBrand::Corona == brand) || (BeerBrand::HoeGaarden == brand);
 }
 
 /**
  * @todo Implement lambda beer country equality comparator
  * @return true if beer county is the same, false otherwise
  */
-auto sameCountry = [](/**???*/)
+auto sameCountry = [](BeerBrand one, BeerBrand two)
```
smay1613, line 40: _const_

PR #295, opened 2021-02-24
algorithms_task_6_Serhii_Starodumov

SStarodumov:


algorithms_task_6/student/include/impl.h
```diff
@@ -18,4 +18,9 @@ void insert_to_sorted(Sequence& target, const FwIt sourceBegin, const FwIt sourc
     static_assert(std::is_const<typename std::remove_reference<
                           typename std::iterator_traits<FwIt>::reference>::type>::value,
                       "FwIt must point to const data");
+    using ValueType = typename FwIt::value_type;
+    std::for_each(sourceBegin, sourceEnd, [&target](const ValueType& value) mutable {
```
SStarodumov, line 22: As I know, it's necessary to lambda be mutable to call non-const method Sequence::insert.... But.. Why it is work without mutable?

PR #298, opened 2021-02-24
algorithms_task_1_verna

VeeM:


algorithms_task_1/student/include/impl.h
```diff
@@ -11,4 +11,21 @@
...
+std::pair<FI, FI> consecutive_group(FI start, FI end, Comparator comp = {})
+{
+    const auto groupStart = std::adjacent_find(start, end, comp);
+
+    if (groupStart == end)
+    {
+        return {end, end};
+    }
+
+    const auto groupValue = *groupStart;
```
smay1613, line 23: what is the reason of copy?

VeeM: oooops :)

algorithms_task_1/student/include/impl.h
```diff
@@ -11,4 +11,21 @@
...
+
+    if (groupStart == end)
+    {
+        return {end, end};
+    }
+
+    const auto& groupValue = *groupStart;
+    auto groupEnd = std::find_if_not(groupStart,
+                                     end,
+                                     [&groupValue, &comp] (decltype(groupValue) value) {
```
smay1613, line 26: prefer traits

PR #325, opened 2021-03-03
algorithms task (2) - Pysarenko

Elzianor:


algorithms_task_2/student/src/impl.cpp
```diff
@@ -1 +1,8 @@
 #include "impl.h"
+
+bool startsWith(const std::string& source, const std::string& prefix)
+{
+    auto result = std::mismatch(std::begin(source), std::end(source), std::begin(prefix));
```
smay1613, line 5: you don't need iterators here

PR #337, opened 2021-03-17
[opichuryn] sequence_task_2

opichuryn:


sequence_task_2/student/include/impl.h
```diff
@@ -14,4 +14,9 @@
  * @note See https://en.cppreference.com/w/cpp/algorithm/rotate
  */
 template<class T, class BI>
-void rotate(std::list<T>& list, BI begin, BI new_begin, BI end);
+void rotate(std::list<T>& list, BI begin, BI new_begin, BI end)
+{
+    std::list<T> tmp;
```
smay1613, line 19: why do you need "tmp" at all?

sequence_task_2/student/include/impl.h
```diff
@@ -14,4 +14,9 @@
...
+    tmp.splice(tmp.end(), list, begin, new_begin);
```
smay1613, line 20: please use only iterators from input

sequence_task_2/student/include/impl.h
opichuryn: Done!

sequence_task_2/student/include/impl.h
opichuryn: Done!

PR #342, opened 2021-03-17
Sequence_task_2_Kosiakova

olsya:


sequence_task_2/student/include/impl.h
```diff
@@ -14,4 +14,7 @@
  * @note See https://en.cppreference.com/w/cpp/algorithm/rotate
  */
 template<class T, class BI>
-void rotate(std::list<T>& list, BI begin, BI new_begin, BI end);
+void rotate(std::list<T>& list, BI begin, BI new_begin, BI end)
+{
+    list.splice(begin, list, new_begin, end);
```
smay1613, line 19: I'd rather move to new_begin position, instead of begin

PR #357, opened 2021-03-20
sequence_task_3 Khudyk

sh17prog:


sequence_task_3/student/include/impl.h
```diff
@@ -8,45 +8,91 @@ class DynamicPlaylist
 {
 public:
     /** @todo Member traits */
+    using size_type = typename Container::size_type;
```
smay1613, line 11: please specify all traits

PR #362, opened 2021-03-22
ordered_associative_task_Serhii_Starodumov

SStarodumov:


ordered_associative_task/student/src/impl.cpp
```diff
@@ -1 +1,6 @@
 #include "impl.h"
+
+size_t convertMarkTo5PointSystem(size_t markIn100Point) {
+    auto range = marksMap.equal_range(markIn100Point);
```
smay1613, line 4: you don't need lower bound

PR #368, opened 2021-03-22
adaptors_task SChyruk

ChNokia:


adaptors_task/student/src/impl.cpp
```diff
@@ -1,8 +1,40 @@
...
+bool isValid(const std::string& source) {
+    std::stack<char> bracketsStack;
+
+    for (const char &ch : source) {
+        auto resultIter = std::find_if(brackets.begin(), brackets.end(), [ch](const std::pair<char, char> value) {
+            return value.first == ch;
+        });
+
+        if (resultIter != brackets.end()) {
+            bracketsStack.push((*resultIter).second);
```
smay1613, line 28: prefer "->"

adaptors_task/student/src/impl.cpp
```diff
@@ -1,8 +1,40 @@
...
+        } else if (isCloseBracket(ch)) {
+            if (bracketsStack.empty()) {
```
smay1613, line 30: you should check stack top value here too

adaptors_task/student/src/impl.cpp
```diff
@@ -1,8 +1,40 @@
...
+    auto resultIter = std::find_if(brackets.begin(), brackets.end(), [bracket](const std::pair<char, char> value) {
+        return value.second == bracket;
+    });
+    return resultIter != brackets.end();
+}
+
+bool isValid(const std::string& source) {
+    std::stack<char> bracketsStack;
+
+    for (const char &ch : source) {
```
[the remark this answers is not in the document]
ChNokia: done

adaptors_task/student/src/impl.cpp
ChNokia: done

adaptors_task/student/src/impl.cpp
ChNokia: I don't understand what do you mean.

smay1613: I mean the case when stack top value (open bracket) has another type

ChNokia: "another type" is it mean other symbol? if other symbol I skip it and add only brackets to the stack or I something don't understand

smay1613: See this example:
[{]}
Stack:
[
{
Then, next symbol is "]" - so, bracket on top of the stack ("}") is not the same type. You can return immediately.

ChNokia: done. thanks

adaptors_task/student/src/impl.cpp
```diff
@@ -1,8 +1,40 @@
...
+            return value.first == symbol;
+        });
+
+        if (resultIter != brackets.end()) {
+            bracketsStack.push(resultIter->second);
+        } else if (isCloseBracket(symbol)) {
+            if (bracketsStack.empty() || bracketsStack.top() != symbol) {
+                return false;
+            }
+            if (bracketsStack.top() == symbol) {
```
smay1613, line 33: you can remove this check

ChNokia: done

PR #374, opened 2021-03-23
[Fedorovskyi] adaptors_task

tra1ning5:


adaptors_task/student/src/impl.cpp
```diff
@@ -1,8 +1,37 @@
...
 };
+
+bool isValid(const std::string &source)
+{
+    std::stack<char> checkBrackets;
+
+    auto isWrongPair = [](const std::map<char, char>& brackets, const char& symbol)
+    {
+        return std::any_of(brackets.begin(), brackets.end(),
+                                         [&symbol](const std::pair<const char, char>& pair) { return symbol == pair.second ? true : false; });
```
smay1613, line 19: Please avoid "true" and "false" here, you don't need them

tra1ning5: Done

PR #377, opened 2021-03-24
Unordered_associative_task_1_Kosiakova

olsya:


unordered_associative_task_1/student/src/impl.cpp
```diff
@@ -1 +1,10 @@
 #include "impl.h"
+#include <unordered_set>
+
+bool is_anagram(const std::string& first, const std::string& second)
+{
+    std::unordered_multiset<char> firstSet(first.begin(), first.end(), first.size());
```
smay1613, line 6: you can also check size first

PR #378, opened 2021-03-24
unordered_task_1_Serhii_Starodumov

SStarodumov:


unordered_associative_task_1/student/src/impl.cpp
```diff
@@ -1 +1,13 @@
 #include "impl.h"
+#include "unordered_set"
+
+bool is_same_content(const playlist &first_playlist, const playlist &second_playlist) {
```
smay1613, line 4: check size first

PR #379, opened 2021-03-24
unordered_associative_task_1 Khudyk

sh17prog:


unordered_associative_task_1/student/src/impl.cpp
```diff
@@ -1 +1,8 @@
 #include "impl.h"
+#include <unordered_set>
+
+bool is_anagram(const std::string& first, const std::string& second) {
+
```
smay1613, line 5: you can also check size first

PR #382, opened 2021-03-25
unordered_associative_task_2_kostiuchenko

kosnko:


unordered_associative_task_2/student/src/impl.cpp
```diff
@@ -41,9 +53,55 @@ filtering_map groupDuplicates (const std::vector<std::string>& dataSource, Filte
     return grouped;
 }
 
+std::vector<std::string> listFiles(const std::string& directory)
+{
+    std::vector<std::string> filePaths;
+    for(const auto& dirEntry : fs::recursive_directory_iterator(directory))
```
smay1613, line 59: you can replace it with std::copy_if

smay1613:
Wow, beautiful!

PR #391, opened 2023-02-02
[Giuseppe Picciuca] functors_task_1

gpicciuca:
All tests pass.

### DIFF
--- a/functors_task_1/student/include/impl.h
+++ b/functors_task_1/student/include/impl.h
@@ -8,6 +8,15 @@
  */
 struct BeerOrganizer
 {
+    int brandId = static_cast<int>(BeerBrand::HoeGaarden);
+
+    BeerBrand operator()()
+    {
+        if (brandId == static_cast<int>(BeerBrand::Max))
+            brandId = static_cast<int>(BeerBrand::HoeGaarden);
+
+        return static_cast<BeerBrand>(brandId++);
+    }
 };
 
 /**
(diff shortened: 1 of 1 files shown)


functors_task_1/student/include/impl.h
```diff
@@ -40,9 +51,17 @@ struct MixingPolicy
      * Whiskey + SevenUp = SevenPlusSeven;
      * Others + Others = Oops;
      */
-    static Cocktail mix(/**???*/)
+    static Cocktail mix(AlcoholDrink alcDrink, NonAlcoholDrink nonAlcDrink)
     {
+        if (alcDrink == AlcoholDrink::Gin && nonAlcDrink == NonAlcoholDrink::LimeJuice)
```
smay1613: _That's a question of style, but prefer always using {} with statements_ https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es85-make-empty-statements-visible

functors_task_1/student/include/impl.h
```diff
@@ -16,16 +25,18 @@ struct BeerOrganizer
...
 {
+    return brand == BeerBrand::Corona || brand == BeerBrand::HoeGaarden;
 }
 
 /**
  * @todo Implement lambda beer country equality comparator
  * @return true if beer county is the same, false otherwise
  */
-auto sameCountry = [](/**???*/)
+auto sameCountry = [](BeerBrand b1, BeerBrand b2)
```
smay1613: _Prefer making params const by default, as modifying function input params is a bad pratice_

functors_task_1/student/include/impl.h
smay1613: _Also, for this task it's okay, but consider following recommendation:_
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es70-prefer-a-switch-statement-to-an-if-statement-when-there-is-a-choice

gpicciuca:
I applied the suggestions you listed above. However, one question arises in the case of the switch-statement in the mix function: Depending on how the compiler is configured, it may complain about missing default-clauses in switches. As in this case, I could have avoided adding the default-clause. Would it been ok or bad-practice?

PR #395, opened 2023-02-02
Functors Task (1) - Icardi 

AlbertoIcardi97:


functors_task_1/student/include/impl.h
```diff
@@ -7,7 +7,15 @@
  * @example HoeGaarden -> Corona -> Carlsberg -> Bud -> ZlataPraha -> Leffe -> HoeGaarden...
  */
 struct BeerOrganizer
-{
+{  
+    BeerBrand operator()(){
+        if(currentIndex==static_cast<int>(BeerBrand::Leffe)){
```
smay1613: Please stick to the "MAX" value

functors_task_1/student/include/impl.h
```diff
@@ -40,9 +51,21 @@ struct MixingPolicy
      * Whiskey + SevenUp = SevenPlusSeven;
      * Others + Others = Oops;
      */
-    static Cocktail mix(/**???*/)
+    static Cocktail mix(const AlcoholDrink AlDrink,const NonAlcoholDrink NonAlDrink)
     {
+        if(AlDrink==AlcoholDrink::Gin && NonAlDrink==NonAlcoholDrink::LimeJuice){
```
smay1613: That's a question of a taste, but consider:
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es70-prefer-a-switch-statement-to-an-if-statement-when-there-is-a-choice

PR #397, opened 2023-02-03
[Giuseppe Picciuca] functors_task_2

gpicciuca:
All tests pass.

### DIFF
--- a/functors_task_2/student/src/impl.cpp
+++ b/functors_task_2/student/src/impl.cpp
@@ -8,47 +8,38 @@ void DataBrowser::userLeave(const std::string &userId)
 bool DataBrowser::getDataType1(const std::string &userId, std::vector<size_t> &returnValues) const
 {
     const auto& it = m_dataReaders.find(userId);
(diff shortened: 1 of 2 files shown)


smay1613:
Please also refactor functions in impl.cpp

functors_task_2/student/src/impl.cpp
```diff
@@ -7,48 +7,54 @@ void DataBrowser::userLeave(const std::string &userId)
...
     {
-        return false;
-    }
-    return it->second->getDataType2(returnValues);
+        auto lambda = [](std::vector<size_t> &returnValues, const IDataSelector* selector) 
+        {
+            return selector->getDataType2(returnValues);
+        };
+
+        auto boundFn = std::bind(lambda, std::placeholders::_2, std::placeholders::_1);
```
smay1613, line 33: Hm, looks interesting if it was intended, but not sure we need to complicate things, arguments order may be changed in lambda itself
But thanks for showing example with bind

PR #403, opened 2023-02-03
[Vincenzo Spinello] functors_task_2

vspinell:


functors_task_2/student/include/impl.h
```diff
@@ -63,8 +64,10 @@ class DataBrowser
     {
         static_assert(std::is_same<decltype(f(nullptr)), bool>::value, "Provided Callable must return bool");
         // find reader
-        // check for errors
-        // call functor
+		auto it = m_dataReaders.find(userId);
```
smay1613, line 67: _formatting issue? check tabs settings_

functors_task_2/student/src/impl.cpp
```diff
@@ -7,48 +7,36 @@ void DataBrowser::userLeave(const std::string &userId)
...
-    if (it->second == nullptr)
-    {
-        return false;
-    }
-    return it->second->getDataType2(returnValues);
+	return safeCall(userId,  [&](const std::unique_ptr<IDataSelector>& a) {
+		auto lambda = [](std::vector<size_t> & returnValues, const IDataSelector* a){
+			return a->getDataType2(returnValues);
+		};
+		auto wrapper = std::bind(lambda, std::placeholders::_2, std::placeholders::_1);
```
smay1613, line 24: we could also change the order inside lambda params
it's ok if std::bind usage was intended

PR #405, opened 2023-02-06
[Iulia Corici] functors_task_2

icorici:


functors_task_2/student/src/impl.cpp
```diff
@@ -7,48 +8,37 @@ void DataBrowser::userLeave(const std::string &userId)
...
-        return false;
-    }
-    if (it->second == nullptr)
-    {
-        return false;
-    }
-    return it->second->getDataType1(returnValues, 0);
+    auto callDataType1 = [this, &returnValues] (const std::unique_ptr<IDataSelector>& selector) {
+        using namespace std::placeholders;
+        return invokeDataRequest(std::bind(&IDataSelector::getDataType1, _1, _2, 0), selector, returnValues);
```
smay1613, line 13: _good during the practice, but lambda should be used in real life_

PR #406, opened 2023-02-06
[Ionut Demsa] functor_task_1

Ionut95:
I mention that I needed to add "size_t" in the "bar_serving.h" file because the compiler complained about it.

### DIFF
--- a/functors_task_1/student/include/impl.h
+++ b/functors_task_1/student/include/impl.h
@@ -25,22 +25,18 @@ public:
  *
  * @note Only Corona and HoeGaarden are expensive
  */
-bool isExpensiveBeer(BeerBrand beer)
+bool isExpensiveBeer(const BeerBrand beer)
 {
-    if (beer == BeerBrand::HoeGaarden || beer == BeerBrand::Corona)
-        return true;
-    return false;
+    return beer == BeerBrand::HoeGaarden || beer == BeerBrand::Corona;
 }
 
 /**
  * @todo Implement lambda beer country equality comparator
  * @return true if beer county is the same, false otherwise
  */
-auto sameCountry = [](BeerBrand beer1, BeerBrand beer2)
+auto sameCountry = [](const BeerBrand beer1, const BeerBrand beer2)
 {
-    if (getBeerCountry(beer1) == getBeerCountry(beer2))
-        return true;
-    return false;
+    return getBeerCountry(beer1) == getBeerCountry(beer2);
 };
 
 struct MixingPolicy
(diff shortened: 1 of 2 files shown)


functors_task_1/student/include/impl.h
```diff
@@ -8,6 +8,15 @@
  */
 struct BeerOrganizer
 {
+private:
+    int next_beer = static_cast<int>(BeerBrand::HoeGaarden);
```
smay1613, line 12: _why don't use none and max?_

functors_task_1/student/include/impl.h
```diff
@@ -16,16 +25,22 @@ struct BeerOrganizer
  *
  * @note Only Corona and HoeGaarden are expensive
  */
-bool isExpensiveBeer(/**???*/)
+bool isExpensiveBeer(BeerBrand beer)
 {
+    if (beer == BeerBrand::HoeGaarden || beer == BeerBrand::Corona)
```
smay1613: usually you should avoid naked true and false in your code, please simplify it, returning the result of the expression

functors_task_1/student/include/impl.h
```diff
@@ -16,16 +25,22 @@ struct BeerOrganizer
...
 }
 
 /**
  * @todo Implement lambda beer country equality comparator
  * @return true if beer county is the same, false otherwise
  */
-auto sameCountry = [](/**???*/)
+auto sameCountry = [](BeerBrand beer1, BeerBrand beer2)
 {
+    if (getBeerCountry(beer1) == getBeerCountry(beer2))
```
smay1613: here too

functors_task_1/student/include/impl.h
```diff
@@ -16,16 +25,22 @@ struct BeerOrganizer
  *
  * @note Only Corona and HoeGaarden are expensive
  */
-bool isExpensiveBeer(/**???*/)
+bool isExpensiveBeer(BeerBrand beer)
```
smay1613: _mark "in" params const by default_

functors_task_1/student/include/impl.h
```diff
@@ -40,9 +55,16 @@ struct MixingPolicy
      * Whiskey + SevenUp = SevenPlusSeven;
      * Others + Others = Oops;
      */
-    static Cocktail mix(/**???*/)
+    static Cocktail mix(AlcoholDrink alcDrink, NonAlcoholDrink nonAlcDrink)
     {
+        if (alcDrink == AlcoholDrink::Gin && nonAlcDrink == NonAlcoholDrink::LimeJuice)
```
smay1613: Also please consider https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es70-prefer-a-switch-statement-to-an-if-statement-when-there-is-a-choice

functors_task_1/student/include/impl.h
```diff
@@ -16,16 +25,22 @@ struct BeerOrganizer
...
+        return true;
+    return false;
 }
 
 /**
  * @todo Implement lambda beer country equality comparator
  * @return true if beer county is the same, false otherwise
  */
-auto sameCountry = [](/**???*/)
+auto sameCountry = [](BeerBrand beer1, BeerBrand beer2)
```
smay1613: mark "in" params const by default, because modifying them is a bad practice

functors_task_1/student/include/impl.h
Ionut95: I thought we are supposed to start with "HoeGarden" as a valid name of brand. I changed "HoeGarden" with "None" and 3 tests failed (Generator.AllEntriesAreValid, Generator.SemiFit and Predicate.DefaultShelf).

smay1613: Yeah, so you should shift the range accordingly, as we use other boundaries

PR #407, opened 2023-02-06
functor_task_1-Darius

DariusAndreiParaschiv:


smay1613:
Please name your PR according to the rules (see README)

functors_task_1/student/include/impl.h
```diff
@@ -1,13 +1,43 @@
...
+
+    public:
+    BeerBrand operator()()
+    {
+        if(index == (int)BeerBrand::Max)
+        {
+            index = 1;
+        }
+
+        switch (index++)
```
smay1613: I think we don't need this switch, because enum is strict ordered, you may just increment it

functors_task_1/student/include/impl.h
```diff
@@ -16,16 +46,18 @@ struct BeerOrganizer
  *
  * @note Only Corona and HoeGaarden are expensive
  */
-bool isExpensiveBeer(/**???*/)
+bool isExpensiveBeer(BeerBrand beer)
```
smay1613: please mark function "in" params const by default, because their modification if considered as a bad practice

functors_task_1/student/include/impl.h
```diff
@@ -16,16 +46,18 @@ struct BeerOrganizer
...
 {
+    return beer == BeerBrand::Corona || beer == BeerBrand::HoeGaarden ? true : false;
```
smay1613: please try to avoid naked true and false in your code, you may directly return result of expression

functors_task_1/student/include/impl.h
```diff
@@ -40,9 +72,17 @@ struct MixingPolicy
      * Whiskey + SevenUp = SevenPlusSeven;
      * Others + Others = Oops;
      */
-    static Cocktail mix(/**???*/)
+    static Cocktail mix(AlcoholDrink alcoholDrink, NonAlcoholDrink nonAlcoholDrink)
     {
+        if(alcoholDrink == AlcoholDrink::Gin && nonAlcoholDrink == NonAlcoholDrink::LimeJuice)
```
smay1613, line 63: Also please consider https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es70-prefer-a-switch-statement-to-an-if-statement-when-there-is-a-choice

functors_task_1/student/include/impl.h
```diff
@@ -1,13 +1,27 @@
...
+
+    public:
+    BeerBrand operator()()
+    {
+        if(index == (int)BeerBrand::Max)
+        {
+            index = 1;
+        }
+
+        return BeerBrand++;
```
smay1613: Does it compile? You incrementing the enum type name

functors_task_1/student/include/impl.h
```diff
@@ -16,16 +30,18 @@ struct BeerOrganizer
...
 }
 
 /**
  * @todo Implement lambda beer country equality comparator
  * @return true if beer county is the same, false otherwise
  */
-auto sameCountry = [](/**???*/)
+auto sameCountry = [](BeerBrand beer1, BeerBrand beer2)
 {
+    return (int)getBeerCountry(beer1) == (int)getBeerCountry(beer2);
```
smay1613: why do you need this cast?

functors_task_1/student/include/impl.h
DariusAndreiParaschiv: forgot to rebuild, working on it

functors_task_1/student/include/impl.h
DariusAndreiParaschiv: I don't remember clearly, but I think it didn't work without the cast

functors_task_1/student/include/impl.h
DariusAndreiParaschiv: I don't understand. What do you mean exactly?

smay1613: Please change param to const BeerBrand - use const by default

functors_task_1/student/include/impl.h
smay1613: try to remove it - this should be safe

PR #408, opened 2023-02-07
[Giuseppe Picciuca] sequence_task_1

gpicciuca:


sequence_task_1/student/include/impl.h
```diff
@@ -8,33 +8,86 @@ class StaticPlaylist
 {
 public:
     /** @todo Member traits */
+    typedef Song_t value_type;
```
smay1613: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rt-using

sequence_task_1/student/include/impl.h
```diff
@@ -8,33 +8,86 @@ class StaticPlaylist
...
+    typedef Song_t& reference;
+    typedef const Song_t& const_reference;
```
smay1613: I'd just stick to the container traits, so in this case we'll introduce a requirement for our class template params
So there is no reason to define them manually

sequence_task_1/student/include/impl.h
```diff
@@ -8,33 +8,86 @@ class StaticPlaylist
...
+    }
 
     /** @todo Assignment from any reversible sequence container */
+    template<class GenericContainer>
+    StaticPlaylist<Container, Song_t>& operator=(const GenericContainer& other)
+    {
+        Container trackList;
+        trackList.assign(other.rbegin(), other.rend());
+
+        m_tracklist.swap(trackList);
```
smay1613: Safe, but I'd rely just on the container assignment

sequence_task_1/student/include/impl.h
```diff
@@ -8,33 +8,86 @@ class StaticPlaylist
...
+    {
+        m_tracklist.emplace(end(), std::forward<Args>(songData)...);
+        return current();
+    }
 
     /** @todo Add track */
-    const Song_t& play(const Song_t& song);
+    const Song_t& play(const Song_t& song)
+    {
+        m_tracklist.emplace(end(), song);
```
smay1613: this concept is closer to insert rather than emplace

sequence_task_1/student/include/impl.h
```diff
@@ -8,33 +8,86 @@ class StaticPlaylist
...
+        return current();
+    }
 
     /** @todo Get first track in playlist stack */
-    const Song_t& current() const;
+    const Song_t& current() const
+    {
+        return *std::prev(end());
```
smay1613: .back()?

sequence_task_1/student/include/impl.h
```diff
@@ -8,33 +8,86 @@ class StaticPlaylist
...
+    }
 
     /** @todo Skip to the next track in playlist, remove current */
-    void switchNext();
+    void switchNext()
+    {
+        if (!hasTracks())
+            throw std::out_of_range("Track list is empty");
+
+        m_tracklist.erase(std::prev(end()));
```
smay1613: pop?

sequence_task_1/student/include/impl.h
```diff
@@ -8,33 +8,86 @@ class StaticPlaylist
...
     /** @todo Member traits */
+    typedef Song_t value_type;
+    typedef Song_t& reference;
+    typedef const Song_t& const_reference;
+    typedef typename Container::iterator iterator;
+    typedef typename Container::const_iterator const_iterator;
+    typedef ptrdiff_t difference_type;
+    typedef size_t size_type;
 
     /** @todo Iterators */
```
smay1613, line 19: what about non-const iterators? (sorry, I should add test for this)

sequence_task_1/student/include/impl.h
```diff
@@ -8,33 +8,86 @@ class StaticPlaylist
...
+        m_tracklist.swap(trackList);
+        return *this;
+    }
 
     /** @todo Add track from initializer */
     template<class... Args>
-    const Song_t& play(Args&&... songData);
+    const Song_t& play(Args&&... songData)
+    {
+        m_tracklist.emplace(end(), std::forward<Args>(songData)...);
```
smay1613: emplace_back?

sequence_task_1/student/include/impl.h
gpicciuca: The test RangeFor.Iterators complaints about the elements being non-const if i add regular non-const iterators. That's why I didn't add them. Even if I mark the begin()/end() methods as const.

smay1613: so, it happens even we have 2 versions of iterators - const and non const?
I'll check and fix the test than, thanks

sequence_task_1/student/include/impl.h
```diff
@@ -8,33 +8,94 @@ class StaticPlaylist
...
+    {
+        m_tracklist.emplace_back(std::forward<Args>(songData)...);
+        return current();
+    }
 
     /** @todo Add track */
-    const Song_t& play(const Song_t& song);
+    const Song_t& play(const Song_t& song)
+    {
+        m_tracklist.insert(m_tracklist.end(), song);
```
smay1613, line 69: push_back?

PR #413, opened 2023-02-07
functor_task_2-Darius

DariusAndreiParaschiv:


functors_task_2/student/include/impl.h
```diff
@@ -65,7 +65,17 @@ class DataBrowser
         // find reader
```
smay1613, line 65: please apply safeCall/invokeDataRequest to the getDataType* functions (you should change them in a way that is accepted by these functions), without changing their purpose (so refactor their code)

PR #417, opened 2023-02-08
[Andrei Tulpan] sequence_task_1

andreitulpan:


sequence_task_1/student/include/impl.h
```diff
@@ -8,33 +8,105 @@ class StaticPlaylist
...
+    {
+        return m_tracklist.crbegin();
+    }
+
+    const_iterator crend() const
+    {
+        return m_tracklist.crend();
+    }
+
+    StaticPlaylist() : m_tracklist() {}
```
smay1613, line 59: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c80-use-default-if-you-have-to-be-explicit-about-using-the-default-semantics

smay1613 (commented):
Excellent!

PR #428, opened 2023-02-09
functor_task_1_Belaidi

BelaidiSiwar:


functors_task_1/student/include/impl.h
```diff
@@ -1,13 +1,51 @@
 #pragma once
 #include "bar_serving.h"
 #include <functional>
+#include <iostream>
+#include <map>
+
+using namespace std;
```
smay1613, line 7: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf6-use-using-namespace-directives-for-transition-for-foundation-libraries-such-as-std-or-within-a-local-scope-only

Please don't!

functors_task_1/student/include/impl.h
```diff
@@ -1,13 +1,51 @@
...
 
 /**
  * @todo Implement functor-generator that will return next beer brand (cyclic)
  * @example HoeGaarden -> Corona -> Carlsberg -> Bud -> ZlataPraha -> Leffe -> HoeGaarden...
  */
+
 struct BeerOrganizer
 {
+    void operator()(){
```
smay1613, line 16: That will not work. We need to return BeerBrand, iterating over the enum, casting it to the int

functors_task_1/student/include/impl.h
```diff
@@ -16,16 +54,18 @@ struct BeerOrganizer
  *
  * @note Only Corona and HoeGaarden are expensive
  */
-bool isExpensiveBeer(/**???*/)
+bool isExpensiveBeer(BeerBrand B)
```
smay1613, line 57: please mark "in" function params const by default

functors_task_1/student/include/impl.h
```diff
@@ -40,9 +80,17 @@ struct MixingPolicy
      * Whiskey + SevenUp = SevenPlusSeven;
      * Others + Others = Oops;
      */
-    static Cocktail mix(/**???*/)
+    static Cocktail mix(AlcoholDrink D1, NonAlcoholDrink D2)
     {
+        static const std::map<std::pair<AlcoholDrink, NonAlcoholDrink>, Cocktail> Cookbook = {
```
smay1613, line 85: Nice solution, but this task we should resolve statically, without containers usage.

PR #429, opened 2023-02-09
sequence_task_2_Belaidi

BelaidiSiwar:


sequence_task_2/student/include/impl.h
```diff
@@ -14,4 +14,9 @@
  * @note See https://en.cppreference.com/w/cpp/algorithm/rotate
  */
 template<class T, class BI>
-void rotate(std::list<T>& list, BI begin, BI new_begin, BI end);
+void rotate(std::list<T>& list, BI begin, BI new_begin, BI end){
+    auto it = begin;
+    while (it != end){
```
smay1613, line 19: you don't need that, use range version of the splice

PR #437, opened 2023-02-10
[Alberto Icardi] - Adaptors_task

AlbertoIcardi97:


adaptors_task/student/src/impl.cpp
```diff
@@ -1,8 +1,41 @@
...
     {'{', '}'}
 };
+
+bool isValid(const std::string& source){
+    std::stack<char> valid;
+
+    for(const char & c : source){
+        
+        for(const auto & b : brackets ){
+            if(c==b.second && valid.empty()){
```
smay1613: please name this condition, it is hard to understand what do you mean here
you may also create a separate function for that task and it will be okay to do search each time for 3 elements

adaptors_task/student/src/impl.cpp
```diff
@@ -1,8 +1,41 @@
...
+                return false;
+            }
+
+            if(c==b.first){
+                valid.push(c);
+                break;
```

try to rewrite and simplify this function, hiding the inner loop in helper

adaptors_task/student/src/impl.cpp
```diff
@@ -1,8 +1,41 @@
...
+            }
+
+            if(c==b.second && !valid.empty() ){
+                auto popped = valid.top();
```
smay1613: try to give more meaningful names for your element
that's a lastOpenBracket or like this

adaptors_task/student/src/impl.cpp
```diff
@@ -1,8 +1,67 @@
...
 
 const std::vector<std::pair<char, char>> brackets {
     {'(', ')'},
     {'[', ']'},
     {'{', '}'}
 };
+
+enum class KindOfChar {OpenBracket, ClosedBracket, NoBracket};
+
+std::pair<KindOfChar,  char> Find_KindOfChar_CorrespondingBracket(const char & c){
```
smay1613, line 14: 1. Please re-consider your naming style - that doesn't match an usual snake case nor camel case
2. Pass literal types by copy

adaptors_task/student/src/impl.cpp
```diff
@@ -1,8 +1,67 @@
...
+         }
+
+         if (c==b.second){
+            return std::make_pair(KindOfChar::ClosedBracket, b.first);
+         }
+
+        
+    }
+
+     return std::make_pair(KindOfChar::NoBracket,'\0');
```
smay1613, line 27: you can just use char {} or like that

adaptors_task/student/src/impl.cpp
```diff
@@ -1,8 +1,67 @@
...
     {'(', ')'},
     {'[', ']'},
     {'{', '}'}
 };
+
+enum class KindOfChar {OpenBracket, ClosedBracket, NoBracket};
+
+std::pair<KindOfChar,  char> Find_KindOfChar_CorrespondingBracket(const char & c){
+    for(const auto & b : brackets ){
+         if(c==b.first){
```
smay1613, line 16: Instead of writing comments, I'd also suggest to add some hints for the reader with help of const bools
if (const isOpenBracket = (sourceChar == bracket_pair.first))

nothing is changed, but condition now have a name

adaptors_task/student/src/impl.cpp
```diff
@@ -1,8 +1,67 @@
...
+bool isValid(const std::string& source){
+
+    std::stack<char> StackOfOpenBrackets;
+
+    for(const char & c : source){
+
+        KindOfChar KindOfc;
+        char CorrespondingBracket;
+
+        std::tie( KindOfc, CorrespondingBracket) = Find_KindOfChar_CorrespondingBracket (c);
```
smay1613, line 40: Cool!

smay1613 (commented):
LGTM

adaptors_task/student/src/impl.cpp
AlbertoIcardi97: Ok, thank you. So, something like this is ok?
std::pair<KindOfChar,  char> find_kindofchar_correspondingbracket(char c)

adaptors_task/student/src/impl.cpp
AlbertoIcardi97: Ok cool!

adaptors_task/student/src/impl.cpp
AlbertoIcardi97: Moreover, how can I choose between snake case and camel case. Is there any situation in which one is better than the other? Can I mix them into a code or should I stay consistent with only one style (e.g only camel or only snake style)?

smay1613: I'd recommend you to choose one style and follow it (if it does not go against your project coding style)
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#nl8-use-a-consistent-naming-style

That's a question of readability and aesthetic
Also consider https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#nl10-prefer-underscore_style-names

AlbertoIcardi97: Ok thank you!

PR #439, opened 2023-02-10
[Carina Ganea] adaptors_task

carina-eliza-ganea:
Modified impl.cpp. All tests pass.

### DIFF
--- a/adaptors_task/student/src/impl.cpp
+++ b/adaptors_task/student/src/impl.cpp
@@ -9,35 +9,49 @@ const std::vector<std::pair<char, char>> brackets {
     {'{', '}'}
 };
 
-struct isBracket
-{
-  isBracket(char _c) : c(_c) {}
-  bool operator()(const std::pair<char,char>& bracketPair) const {
-    return c == bracketPair.first || c == bracketPair.second;
-  }
-  private:
-    char c;
-};
+bool add_or_fail(std::stack<char>& openBrackets, char& c ){
+    auto is_opening_bracket = [c] (const std::pair<char,char>& b) { return c == b.first;};
+    auto is_closing_bracket = [c] (const std::pair<char,char>& b) { return c == b.second;};
+
+    if(std::find_if(brackets.begin(), brackets.end(),is_closing_bracket)!= brackets.end()){
+        return false;
+    }
+
+    if(std::find_if(brackets.begin(), brackets.end(),is_opening_bracket)!= brackets.end()){
+        openBrackets.push(c);
+    }
+
+    return true;
+}
 
 bool isValid(const std::string& source){
     std::stack<char> openBrackets;
 
     for(auto c : source){
-        const bool closeable_with_current_character = !openBrackets.empty() && 
-        std::find(brackets.begin(), brackets.end(), std::make_pair(openBrackets.top(), c)) != brackets.end();
 
-        if(openBrackets.empty() || !closeable_with_current_character){
-            
-            if(std::find_if(brackets.begin(), brackets.end(), isBracket(c)) != brackets.end()){
-                openBrackets.push(c);
+        if( !openBrackets.empty()){
+            auto bracket_pair = std::find(brackets.begin(), brackets.end(), std::make_pair(openBrackets.top(), c));
+
+            const bool closeable_with_current_character = bracket_pair != brackets.end();
+
+            if(closeable_with_current_character){
+                
+                openBrackets.pop();
+
+            } else {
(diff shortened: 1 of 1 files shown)


adaptors_task/student/src/impl.cpp
```diff
@@ -1,8 +1,37 @@
...
 const std::vector<std::pair<char, char>> brackets {
     {'(', ')'},
     {'[', ']'},
     {'{', '}'}
 };
+
+bool isValid(const std::string& source){
+    std::stack<char> notClosed;
+
+    for(auto i = 0; i < source.size(); i++){
```
smay1613: Always prefer range for loops!

adaptors_task/student/src/impl.cpp
```diff
@@ -1,8 +1,37 @@
...
+#include <stack>
+#include <iostream>
 const std::vector<std::pair<char, char>> brackets {
     {'(', ')'},
     {'[', ']'},
     {'{', '}'}
 };
+
+bool isValid(const std::string& source){
+    std::stack<char> notClosed;
```
smay1613: _openBrackets?_

adaptors_task/student/src/impl.cpp
```diff
@@ -1,8 +1,37 @@
...
+
+    for(auto i = 0; i < source.size(); i++){
+        bool closed = false;
+
+        if(!notClosed.empty()){
+            for(auto b : brackets){
```
smay1613: 1. Use named algorithm (like find_if)
2. Create new separate named function

adaptors_task/student/src/impl.cpp
```diff
@@ -1,8 +1,37 @@
...
+                    notClosed.pop();
+                    closed = true;
+                    break;
+                }  
+            }
+        }
+        
+        if(!closed){
+            for( auto b : brackets){
+                if(source[i] == b.first || source[i] == b.second){
```
smay1613: why should we check for closed brackets here, if we're in "non closed" branch?

smay1613: Please always give name for your conditions of such type, use const bool variables

adaptors_task/student/src/impl.cpp
```diff
@@ -1,8 +1,37 @@
...
+
+    for(auto i = 0; i < source.size(); i++){
+        bool closed = false;
+
+        if(!notClosed.empty()){
+            for(auto b : brackets){
+                if(std::make_pair(notClosed.top(), source[i]) == b){
+                    notClosed.pop();
+                    closed = true;
+                    break;
```

adaptors_task/student/src/impl.cpp
carina-eliza-ganea: This was to check if the current character is a bracket and not some other character. It will not be added to the stack unless it's a bracket. Logically yes, I wouldn't need to add a closed bracket to the stack since it would mean it is an invalid string, but my invalid condition is whether the stack is empty or not at the end.

adaptors_task/student/src/impl.cpp
```diff
@@ -1,8 +1,44 @@
...
 const std::vector<std::pair<char, char>> brackets {
     {'(', ')'},
     {'[', ']'},
     {'{', '}'}
 };
+
+struct isBracket
+{
+  isBracket(char _c) : c(_c) {}
+  bool operator()(const std::pair<char,char>& bracketPair) const {
```
smay1613: no sense to wrap it in functor, use free function instead

adaptors_task/student/src/impl.cpp
```diff
@@ -1,8 +1,44 @@
...
+  private:
+    char c;
+};
+
+bool isValid(const std::string& source){
+    std::stack<char> openBrackets;
+
+    for(auto c : source){
+        const bool closeable_with_current_character = !openBrackets.empty() && 
+        std::find(brackets.begin(), brackets.end(), std::make_pair(openBrackets.top(), c)) != brackets.end();
```
smay1613: please name the result of search also, like const bool closedBracketMatching

adaptors_task/student/src/impl.cpp
```diff
@@ -1,8 +1,44 @@
...
+
+        if(openBrackets.empty() || !closeable_with_current_character){
+            
+            if(std::find_if(brackets.begin(), brackets.end(), isBracket(c)) != brackets.end()){
+                openBrackets.push(c);
+            }
+
+        } else {
```
smay1613: what about early exit when we find that closing bracket does not match the opened?

PR #440, opened 2023-02-10
[Iulia Corici] sequence_task_1

icorici:


sequence_task_1/student/include/impl.h
```diff
@@ -8,33 +8,77 @@ class StaticPlaylist
...
+    const_iterator begin() const {
+        return m_tracklist.cbegin();
+    }
 
-    StaticPlaylist() = default;
+    const_iterator end() const {
+        return m_tracklist.cend();
+    }
+
+    StaticPlaylist() {
```
smay1613, line 29: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c80-use-default-if-you-have-to-be-explicit-about-using-the-default-semantics

sequence_task_1/student/include/impl.h
```diff
@@ -8,33 +8,77 @@ class StaticPlaylist
...
+        m_tracklist= Container();
+    }
 
     /** @todo Constructor from any reversible sequence container */
+    template<class OtherContainer>
+    StaticPlaylist(const OtherContainer &container){
+        m_tracklist = Container(container.rbegin(), container.rend());
```
smay1613, line 36: Use initialization list for this task, omit the typename and perform direct initialization, rather than assignment

sequence_task_1/student/include/impl.h
```diff
@@ -8,33 +8,77 @@ class StaticPlaylist
...
-    const Song_t& play(Args&&... songData);
+    const Song_t& play(Args&&... songData) {
+        m_tracklist.emplace_back(std::forward<Args>(songData)...);
+        return current();
+    }
 
     /** @todo Add track */
-    const Song_t& play(const Song_t& song);
+    const Song_t& play(const Song_t& song) {
+        m_tracklist.emplace_back(song);
```
smay1613, line 55: This concept is much closer to push_back, as we have the lvalue song

PR #444, opened 2023-02-10
[Bogdan Pana] adaptors_task

BogdanPana26:


adaptors_task/student/src/impl.cpp
```diff
@@ -1,8 +1,43 @@
...
+bool isValid(const std::string& source)
+{
+    std::stack<char> br_stack;
+
+    for(const char &ch : source)
+    {
+
+        for(const auto &bracket : brackets )
+        {
+            if(br_stack.empty() && ch==bracket.second )
```
smay1613: please give this condition a name, what ch==bracket.second means

adaptors_task/student/src/impl.cpp
```diff
@@ -1,8 +1,43 @@
...
+            {
+                return false;
+            }
+
+            if(ch==bracket.first)
+            {
+                br_stack.push(ch);
+                break;
```

adaptors_task/student/src/impl.cpp
```diff
@@ -1,8 +1,43 @@
...
+            }
+
+            if(!br_stack.empty() && ch==bracket.second)
+            {
+                auto first = br_stack.top();
```
smay1613: _lastOpened?_

adaptors_task/student/src/impl.cpp
```diff
@@ -12,26 +12,30 @@ bool isValid(const std::string& source)
...
 
         for(const auto &bracket : brackets )
         {
-            if(br_stack.empty() && ch==bracket.second )
+            if(br_stack.empty() && isBracketClosed(ch,bracket))
             {
                 return false;
             }
 
             if(ch==bracket.first)
```
smay1613, line 30: and what about this condition?

adaptors_task/student/src/impl.cpp
```diff
@@ -12,26 +12,30 @@ bool isValid(const std::string& source)
 {
     std::stack<char> br_stack;
 
+    auto isBracketClosed = [](const char ch ,const std::pair<char, char> bracket)
```
smay1613, line 15: so maybe we'll pack the loop into the lambda?

PR #446, opened 2023-02-12
[Andrei Tulpan] adaptors_task

andreitulpan:


adaptors_task/student/include/impl.h
```diff
@@ -11,4 +12,38 @@
  *
  * @return true if expression is valid
  */
-bool isValid(const std::string& source);
+bool isValid(const std::string& source) {
+    std::stack<char> stack;
+    for (auto const &c : source) {
+        if (c == '(' || c == '[' || c == '{') {
```
smay1613: It's better to pack this into some container

adaptors_task/student/include/impl.h
```diff
@@ -11,4 +12,38 @@
...
+            stack.push(c);
+        } else if (c == ')' || c == ']' || c == '}') {
+            if (stack.empty()) {
+                return false;
+            }
+            char ch = stack.top();
+            stack.pop();
+            switch (ch) {
```
smay1613: this should be definitely a search in the container

adaptors_task/student/include/impl.h
```diff
@@ -11,4 +12,38 @@
...
+                    break;
+                case '{':
+                    if (c != '}') {
+                        return false;
+                    }
+                    break;
+            }
+        }
+    }
+    if (!stack.empty()) {
```
smay1613: please avoid naked true and false, just return stack.empty()

adaptors_task/student/include/impl.h
```diff
@@ -11,4 +12,38 @@
  *
  * @return true if expression is valid
  */
-bool isValid(const std::string& source);
+bool isValid(const std::string& source) {
+    std::stack<char> stack;
```
smay1613: Please express the intent, what is the "stack"? Maybe openedBrackets?

andreitulpan: Yes, I wasn't to suggestive with this name. I resolved that.

adaptors_task/student/include/impl.h
andreitulpan: done

adaptors_task/student/include/impl.h
andreitulpan: I did it and the code is smaller now and can be extended much more simple.

adaptors_task/student/include/impl.h
andreitulpan: done

adaptors_task/student/src/impl.cpp
```diff
@@ -6,3 +6,28 @@ const std::vector<std::pair<char, char>> brackets {
     {'[', ']'},
     {'{', '}'}
 };
+
+bool isValid(const std::string& source) {
+    std::stack<char> openedBrackets;
+    for (auto const &c : source) {
+        for (auto const &currentBracket : brackets) {
+            if (c == currentBracket.first) {
```
smay1613: be more explicit like:
if (const bool isOpenBracket = (c == currentBracket.first))

PR #447, opened 2023-02-13
[Giuseppe Picciuca] ordered_associative_task

gpicciuca:


ordered_associative_task/student/include/impl.h
```diff
@@ -15,4 +23,10 @@
  *
  * @return 1-5
  */
-size_t convertMarkTo5PointSystem(size_t markIn100Point);
+size_t convertMarkTo5PointSystem(size_t markIn100Point)
+{
+    const size_t MAX_POINTS = 100;
+    auto it = gradePoints.lower_bound(std::min<size_t>(markIn100Point, MAX_POINTS));
```
smay1613, line 29: that's a way, but in general we should go on previous iterator

PR #449, opened 2023-02-13
Sequence task 1-Darius

DariusAndreiParaschiv:


functors_task_2/student/include/impl.h
```diff
@@ -65,7 +65,17 @@ class DataBrowser
         // find reader
         // check for errors
         // call functor
-        return false;
+        const auto& it = m_dataReaders.find(userId);
```
smay1613, line 68: Please start each new task from the master branch, don't include this into your next prs!

sequence_task_1/student/include/impl.h
```diff
@@ -3,38 +3,85 @@
...
-    StaticPlaylist() = default;
+    const_iterator end() const {
+        return m_tracklist.cend();
+    }
 
+    StaticPlaylist() : m_tracklist() {}
     /** @todo Constructor from any reversible sequence container */
-
+    template<class PlayList>
+    StaticPlaylist(PlayList& playList)
```
smay1613: 1. You don't modify the playlist, so you should not take the non-const reference to it
2. You may use the constructor with two iterators in the initialization list

PR #450, opened 2023-02-13
Sequence task 2-Darius

DariusAndreiParaschiv:


smay1613:
Please start each new task from the master, so PR will not include previous tasks

PR #451, opened 2023-02-13
Sequence task 3-Darius

DariusAndreiParaschiv:


sequence_task_3/student/include/impl.h
```diff
@@ -8,45 +8,100 @@ class DynamicPlaylist
...
+    const_iterator end() const {
+        return m_tracklist.cend();
+    }
 
     DynamicPlaylist() = default;
 
     /** @todo Constructor from two iterators*/
-
+    template<typename iterator>
+    DynamicPlaylist(iterator first, iterator last)
```
smay1613: use constructor directly and initialize instead of assignment

sequence_task_3/student/include/impl.h
```diff
@@ -8,45 +8,100 @@ class DynamicPlaylist
...
+    {
+        m_tracklist.assign(first, last);
+    };
     /** @todo Set track as currently played from initializer */
     template<class... Args>
-    const Song_t& play(Args&&... songData);
+    const Song_t& play(Args&&... songData)
+    {
+        m_tracklist.emplace(begin(), std::forward<Args>(songData)...);
```
smay1613: emplace_front

sequence_task_3/student/include/impl.h
```diff
@@ -8,45 +8,100 @@ class DynamicPlaylist
...
+        return current();
+    };
 
     /** @todo Set track as currently played from lvalue */
-    const Song_t& play(const Song_t& song);
+    const Song_t& play(const Song_t& song)
+    {
+        return m_tracklist.insert(begin(), song);
```
smay1613: push_front

sequence_task_3/student/include/impl.h
```diff
@@ -8,45 +8,100 @@ class DynamicPlaylist
...
+        return current();
+    };
 
     /** @todo Set track to be played next from initializer */
     template<class... Args>
-    void playNext(Args&&... songData);
+    void playNext(Args&&... songData)
+    {
+        int pos = hasTracks() ? 1:0;
```
smay1613: looks hacky.
if there is no tracks in playlist, we should play, in other case - insert

sequence_task_3/student/include/impl.h
```diff
@@ -8,45 +8,100 @@ class DynamicPlaylist
...
+    void add(Args&&... songData)
+    {
+        m_tracklist.emplace_back(std::forward<Args>(songData)...);
+    };
 
     /** @todo Add track to the end of the queue from lvalue */
-    void add(const Song_t& song);
+    void add(const Song_t& song)
+    {
+        m_tracklist.insert(end(), song);
```
smay1613: push_back?

sequence_task_3/student/include/impl.h
DariusAndreiParaschiv: Can you explain to me in more detail, please?

sequence_task_3/student/include/impl.h
DariusAndreiParaschiv: What exactly should I do? I don't understand

sequence_task_3/student/include/impl.h
smay1613: https://www.learncpp.com/cpp-tutorial/constructor-member-initializer-lists/

Here, you are doing two operations:
m_tracklist {}
m_tracklist.assign(first,last)

Why can't we merge whem?

sequence_task_3/student/include/impl.h
smay1613: I'd do like
if (!hasTracks)
{
play(songData)
}
else
{
// insertion to the next
}

sequence_task_3/student/include/impl.h
DariusAndreiParaschiv: I understand the initialization, but how am I able to use the iterators afterwards?

smay1613: You can use constructor that accepts two iterators: m_tracklist {first, last};

PR #452, opened 2023-02-13
Adaptors task-Darius

DariusAndreiParaschiv:


adaptors_task/student/include/impl.h
```diff
@@ -11,4 +11,52 @@
  *
  * @return true if expression is valid
  */
-bool isValid(const std::string& source);
+
+bool isValid(const std::string& source)
+{
+    std::stack<char> brackets;
+    std::string close = ")]}";
```
smay1613: reuse them from from impl.cpp

adaptors_task/student/include/impl.h
```diff
@@ -11,4 +11,52 @@
...
+    std::string open = "([{";
+    for(int i = 0; i < source.size(); i++)
```
smay1613: avoid index-for, always use range for!

adaptors_task/student/include/impl.h
```diff
@@ -11,4 +11,52 @@
...
+        }
+        else if(close.find(source[i]) != std::string::npos)
+        {
+            if(brackets.empty())
+                return false;
+            switch(source[i])
+            {
+                case ')':
+                {
+                    if(brackets.top() != '(')
```
smay1613: so if we add new type of brackets, we will modify the code everywhere again?

adaptors_task/student/src/impl.cpp
```diff
@@ -1,8 +1,48 @@
...
+                        return false;
+                    else
+                        bracketsStack.pop();
+                }
+            }
+        }
+    }
+    if(!bracketsStack.empty())
+        return false;
+    return true;
```
smay1613, line 47: return bracketsStack.empty()

adaptors_task/student/src/impl.cpp
```diff
@@ -1,8 +1,48 @@
...
+        }
+        else if(close.find(character) != std::string::npos)
+        {
+            if(bracketsStack.empty())
+                return false;
+            for(auto bracket : brackets)
+            {
+                if(bracket.second == character)
+                {
+                    if(bracketsStack.top() != bracket.first)
```
smay1613, line 37: please be more explicit with your conditions.
use const bool for that - if (const bool isOpened = ...)/if (const bool bracketsMatch = ..) 
that may improve code readability

PR #453, opened 2023-02-13
[Carina Ganea] ordered_associative_task

carina-eliza-ganea:
All tests pass

### DIFF
--- a/ordered_associative_task/student/include/impl.h
+++ b/ordered_associative_task/student/include/impl.h
@@ -16,17 +16,20 @@
  * @return 1-5
  */
 
-const size_t MAX_POINTS = 100;
-
 std::map<std::size_t, std::size_t> pointSystem{
-    {29,1},
-    {49,2},
-    {69,3},
-    {89,4},
-    {100, 5}
+    {0,1},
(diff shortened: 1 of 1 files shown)


ordered_associative_task/student/include/impl.h
```diff
@@ -15,4 +15,18 @@
  *
  * @return 1-5
  */
-size_t convertMarkTo5PointSystem(size_t markIn100Point);
+
+const size_t MAX_POINTS = 100;
```
smay1613: hacky way, I'd rely on such mapping only:

ordered_associative_task/student/include/impl.h
```diff
@@ -16,17 +16,20 @@
...
-    {100, 5}
+    {0,1},
+    {30,2},
+    {50,3},
+    {70,4},
+    {90, 5}
 };
 
 std::size_t convertMarkTo5PointSystem(std::size_t markIn100Point){
+    auto res = pointSystem.lower_bound(markIn100Point);
```
smay1613, line 28: and what if we just try upper_bound and then go back?

PR #455, opened 2023-02-13
adaptors_task_Belaidi

BelaidiSiwar:


adaptors_task/student/include/impl.h
```diff
@@ -1,5 +1,8 @@
 #pragma once
 #include <string>
+#include <stack>
+
+using namespace std;
```
smay1613, line 5: No no no no!
https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf7-dont-write-using-namespace-at-global-scope-in-a-header-file

adaptors_task/student/include/impl.h
```diff
@@ -11,4 +14,42 @@
  *
  * @return true if expression is valid
  */
-bool isValid(const std::string& source);
+bool isValid(const std::string& source){
+    stack <string> openingbrackets;
+    for (const char ch : source){
+        switch (ch)
+        {
+        case '(':
```
smay1613, line 22: please reuse this info from impl.cpp

adaptors_task/student/include/impl.h
```diff
@@ -11,4 +14,42 @@
  *
  * @return true if expression is valid
  */
-bool isValid(const std::string& source);
+bool isValid(const std::string& source){
+    stack <string> openingbrackets;
```
smay1613, line 18: why do you store the string and not the char?

adaptors_task/student/include/impl.h
```diff
@@ -11,4 +14,42 @@
...
+        
+        case '}':
+            if (openingbrackets.empty() || openingbrackets.top() != "{"){
+                return false;
+            }
+            openingbrackets.pop();
+            break;
+        
+        case ']':
+            if (openingbrackets.empty() || openingbrackets.top() != "["){
```
smay1613, line 43: we'll need to repeat this each time we introduce/change a bracket type, so please let's store this info in container from impl.cpp

PR #456, opened 2023-02-13
[Andrei Tulpan] ordered_associative_task

andreitulpan:


ordered_associative_task/student/include/impl.h
```diff
@@ -15,4 +16,15 @@
...
+std::map<size_t, size_t> markLimits {
+    {29, 1},
+    {49, 2},
+    {69, 3},
+    {89, 4},
+    {100, 5}
+};
+
+size_t convertMarkTo5PointSystem(size_t markIn100Point) {
+    return markLimits.lower_bound(std::min(markIn100Point, (size_t)100))->second;
```
smay1613: hacky way, try to solve it with such mapping:

Also, why should you cast 100?
use for constants https://en.cppreference.com/w/cpp/language/integer_literal

andreitulpan: I casted 100 because std::min need arguments of the same type. I resolved the both problems.

ordered_associative_task/student/include/impl.h
```diff
@@ -17,14 +17,17 @@
...
+    {0, 1},
+    {30, 2},
+    {50, 3},
+    {70, 4},
+    {90, 5}
 };
 
 size_t convertMarkTo5PointSystem(size_t markIn100Point) {
-    return markLimits.lower_bound(std::min(markIn100Point, (size_t)100))->second;
+    auto iterator = markLimits.upper_bound(std::min(markIn100Point, maxScore));
```
smay1613: you don't need maxScore here
it will return end iterator that will be safely iterated back
and we'll not get under the "0"

andreitulpan: you're right. I got it, thanks.

PR #457, opened 2023-02-13
Sequence_task_1_ApostolRazvan

ApostolRazvan:


sequence_task_1/student/include/impl.h
```diff
@@ -8,33 +8,67 @@ class StaticPlaylist
...
+    }
+    const_iterator end() const{
+        return m_tracklist.cend();
+    }
 
     StaticPlaylist() = default;
 
     /** @todo Constructor from any reversible sequence container */
-
+    template<class AuxContainer> StaticPlaylist(const AuxContainer &aux){
```
smay1613, line 29: You may use sequence constuctor here in initialization list

functors_task_1/student/include/impl.h
```diff
@@ -8,6 +8,15 @@
  */
 struct BeerOrganizer
 {
+private:
```
smay1613, line 11: Please never commit previous tasks in the new one!
Start each task from the master branch

PR #458, opened 2023-02-13
[Vincenzo Spinello] ordered_associative_task

vspinell:


ordered_associative_task/student/src/impl.cpp
```diff
@@ -1 +1,16 @@
 #include "impl.h"
+#include <map>
+
+size_t convertMarkTo5PointSystem(size_t markIn100Point) {
+	const size_t MAX = 100;
```
smay1613: hacky way, try to solve it with such mapping:

ordered_associative_task/student/src/impl.cpp
```diff
@@ -2,15 +2,11 @@
...
-		{49, 2},
-		{69, 3},
-		{89, 4},
-		{MAX, 5}
+		{0, 1}, {30, 2}, {50, 3}, {70, 4}, {90, 5}
 	};
-	auto nearestRange = ranges.lower_bound(markIn100Point);
-	const bool isOutOfBound = nearestRange == ranges.end();
-	return isOutOfBound? ranges[MAX] : nearestRange->second;
+	auto nearestRange = ranges.equal_range(markIn100Point);
```
smay1613, line 8: simpler - we could use upper bound and safely go back

PR #459, opened 2023-02-13
[Claros Alexis] ordered_associative_task

AlexLUXgit:


ordered_associative_task/student/include/impl.h
```diff
@@ -15,4 +15,8 @@
  *
  * @return 1-5
  */
-size_t convertMarkTo5PointSystem(size_t markIn100Point);
+int convertMarkTo5PointSystem(int markIn100Point){
+    if(markIn100Point>90) markIn100Point = 90;
```
smay1613, line 19: hacky way, try to solve it with such mapping:

PR #461, opened 2023-02-13
[Alberto Icardi] -Ordered associative task 

AlbertoIcardi97:


ordered_associative_task/student/src/impl.cpp
```diff
@@ -1 +1,6 @@
 #include "impl.h"
+std::map<size_t,size_t> mapConverter={{29,1},{49,2},{69,3},{89,4},{SIZE_MAX,5}};
```
smay1613: Where SIZE_MAX is defined?

ordered_associative_task/student/include/impl.h
```diff
@@ -1,6 +1,8 @@
 #pragma once
 #include <map>
 
+extern std::map<size_t,size_t> mapConverter;
```
smay1613: why do you need extern here?

ordered_associative_task/student/src/impl.cpp
```diff
@@ -1 +1,6 @@
...
+
+size_t convertMarkTo5PointSystem(size_t markIn100Point){
+     return mapConverter.lower_bound(markIn100Point)->second;
```
smay1613: what if lower_bound will return end iterator?

ordered_associative_task/student/src/impl.cpp
AlbertoIcardi97: Is not SIZE_MAX the maximum value of size_t?
https://en.cppreference.com/w/c/types/limits

ordered_associative_task/student/include/impl.h
AlbertoIcardi97: I used extern in order to declare the map in the header and define it in the source. Actually, I can declare and define it only in the cpp file, right?

ordered_associative_task/student/src/impl.cpp
AlbertoIcardi97: Since you cannot pass a size_t value greater than size_max, I thought that whatever value you pass to lower_bound never returns an end iterator

ordered_associative_task/student/include/impl.h
smay1613: yes

ordered_associative_task/student/src/impl.cpp
smay1613: hacky way, I'd rely on such mapping only:

ordered_associative_task/student/src/impl.cpp
```diff
@@ -1,6 +1,14 @@
 #include "impl.h"
-std::map<size_t,size_t> mapConverter={{29,1},{49,2},{69,3},{89,4},{SIZE_MAX,5}};
+
+std::map<size_t,size_t> mapConverter={{0,1},{30,2},{50,3},{70,4},{90,5}};
 
 size_t convertMarkTo5PointSystem(size_t markIn100Point){
-     return mapConverter.lower_bound(markIn100Point)->second;
+    auto result=mapConverter.lower_bound(markIn100Point);
+
+    if(result==mapConverter.end() || result->first>markIn100Point){
```
smay1613, line 8: with upper_bound you may not check for the end

PR #462, opened 2023-02-13
[Bogdan Pana] ordered_associative_task

BogdanPana26:


ordered_associative_task/student/include/impl.h
```diff
@@ -15,4 +15,22 @@
  *
  * @return 1-5
  */
-size_t convertMarkTo5PointSystem(size_t markIn100Point);
+std::size_t convertMarkTo5PointSystem(std::size_t markIn100Point)
+{
+    if(markIn100Point > 100)
```
smay1613: Looks hacky, try to solve it with such mapping:

ordered_associative_task/student/include/impl.h
```diff
@@ -17,20 +17,13 @@
...
+        {70,4},
+        {90,5}
     };
-
-    auto val = MarkDict.lower_bound(markIn100Point);
-    return MarkDict[val->first];
-
-
+    auto val = MarkDict.upper_bound(markIn100Point);
+    return MarkDict[(--val)->first];
```
smay1613, line 28: you can just return --val->second

PR #463, opened 2023-02-13
[Ionut Demsa] adaptors_task

Ionut95:


adaptors_task/student/src/impl.cpp
```diff
@@ -1,8 +1,43 @@
...
     {'(', ')'},
     {'[', ']'},
     {'{', '}'}
 };
+
+bool isValid(const std::string &source) {
+    std::stack<char> notClosedBr;
+    for (char c : source) {
+        switch(c) {
+            case '(':
```
smay1613, line 15: You should reuse this info from impl.cpp

adaptors_task/student/src/impl.cpp
```diff
@@ -1,8 +1,43 @@
...
+#include <stack>
 
 const std::vector<std::pair<char, char>> brackets {
     {'(', ')'},
     {'[', ']'},
     {'{', '}'}
 };
+
+bool isValid(const std::string &source) {
+    std::stack<char> notClosedBr;
```
smay1613, line 12: so, it means _openingBrackets_?

adaptors_task/student/src/impl.cpp
```diff
@@ -1,8 +1,43 @@
...
+                notClosedBr.pop();
+                break;
+            case ']':
+                if(notClosedBr.empty() || notClosedBr.top() != brackets.at(1).first) {
+                    return false;
+                }
+                notClosedBr.pop();
+                break;
+            case '}':
+                if (notClosedBr.empty() || notClosedBr.top() != brackets.at(2).first) {
```
smay1613, line 33: hacky and not maintainable
please use search in brackets container from impl.cpp

PR #469, opened 2023-02-14
[Alberto Icardi] - Unordered Associative Task 1 

AlbertoIcardi97:


unordered_associative_task_1/student/include/impl.h
```diff
@@ -32,4 +48,9 @@ using playlist = std::vector<Song>;
  *
  * @return true if first playlist is anagram of second
  */
-bool is_same_content(const playlist& first_playlist, const playlist& second_playlist);
+bool is_same_content(const playlist& first_playlist, const playlist& second_playlist){
+    std::unordered_multiset<Song> first_multiset (first_playlist.begin(),first_playlist.end());
+    std::unordered_multiset<Song> second_multiset (second_playlist.begin(),second_playlist.end());
+    return first_multiset==second_multiset;
```
smay1613, line 54: so what is the profit of using std::unordered_multiset here? what is the difference with comparing vectors?

smay1613: looks too good to be true, so please add some explanation to these lines

AlbertoIcardi97: The difference with comparing vector is that the comparison between vectors take into account also the order, while the one between multisets no. 
The explanation of the code is that I implement a multiset of Song, so I custom the hash function, and I need also to overload the ==operator.
At this point, I can use the range constructor to define 2 multisets starting from the elements contained in the vectors first_playlist and second_playlist.
Finally, I can achieve check if the two playlist has the same content by using the ==operator between the two multisets. 
Why wouldn't it work?

smay1613: Yeah, thanks, I agree with that, that's a best solution for this task, just wanted to be sure you understand how == will work for unordered multiset.
Also I'd recommend to check sizes before copying.

AlbertoIcardi97: Okay :') Thank you!

PR #470, opened 2023-02-14
[Vincenzo Spinello] unordered_associative_task_1

vspinell:


vspinell:
I wanted to avoid the overload of the operator== of the defined unordered_set type but I occurred into this error:

"no match for ‘operator==’ (operand types are ‘const Song’ and ‘const Song’)
  115 |       { return *__it1 == *__it2; }
      |                ~~~~~~~^~~~~~~~~"
      
Is there a way to solve this issue without using the first overload?

unordered_associative_task_1/student/src/impl.cpp
```diff
@@ -1 +1,43 @@
...
+		return s1.name() == s2.name();
+	}
+};
+
+template<typename T>
+const bool isEqualSize(const T& p1, const T& p2) {
+	return p1.size() == p2.size();
+}
+
+using unordered_set = std::unordered_set<Song, SongHash, SongEqual>;
```
smay1613: please don't, write the full type

unordered_associative_task_1/student/src/impl.cpp
```diff
@@ -1 +1,43 @@
 #include "impl.h"
+#include <unordered_set>
+
+struct SongHash {
+	std::size_t operator ()(const Song& s) const {
+		return std::hash<std::string>{}(s.name());
+	}
+};
+
+struct SongEqual {
```
smay1613: it's better to provide ==, don't wrap it in functor please for this task

unordered_associative_task_1/student/src/impl.cpp
```diff
@@ -1 +1,43 @@
 #include "impl.h"
+#include <unordered_set>
+
+struct SongHash {
```
smay1613: now try to extend std hash instead of using functor

unordered_associative_task_1/student/src/impl.cpp
```diff
@@ -1 +1,43 @@
...
+}
+
+using unordered_set = std::unordered_set<Song, SongHash, SongEqual>;
+
+const bool operator== (const unordered_set& p1, const unordered_set& p2) {
+	if (!isEqualSize(p1, p2)) {
+		return false;
+	}
+	for (auto &song : p2) {
+		if (p1.count(song) != 1){
```
smay1613: count may be implicitly casted to a bool, so you may just omit it and use !count

unordered_associative_task_1/student/src/impl.cpp
```diff
@@ -1 +1,43 @@
...
+
+template<typename T>
+const bool isEqualSize(const T& p1, const T& p2) {
+	return p1.size() == p2.size();
+}
+
+using unordered_set = std::unordered_set<Song, SongHash, SongEqual>;
+
+const bool operator== (const unordered_set& p1, const unordered_set& p2) {
+	if (!isEqualSize(p1, p2)) {
```
smay1613: not sure we need to create a separate function for that

unordered_associative_task_1/student/src/impl.cpp
```diff
@@ -1 +1,43 @@
...
+}
+
+
+bool is_same_content(const playlist& first_playlist, const playlist& second_playlist) {
+	if (!isEqualSize(first_playlist, second_playlist)){
+		return false;
+	}
+	unordered_set playlist1(first_playlist.begin(), first_playlist.end());
+	unordered_set playlist2(second_playlist.begin(), second_playlist.end());
+	return playlist1 == playlist2;
```
smay1613, line 21: will it work for following case?
{"1", "1", "2"}
{"2", "2", "1"}

unique songs are "1" and "2", size is equal, but the amount of tracks differs

unordered_associative_task_1/student/src/impl.cpp
```diff
@@ -1 +1,22 @@
...
+		}
+	};
+}
+
+bool operator==(const Song& s1, const Song& s2) {
+	return s1.name() == s2.name();
+}
+
+bool is_same_content(const playlist& first_playlist, const playlist& second_playlist) {
+	std::unordered_multiset<Song> playlist1(first_playlist.begin(), first_playlist.end());
```
smay1613, line 19: also you may check the size before copying

PR #473, opened 2023-02-14
[Giuseppe Picciuca] unordered_associative_task_2

gpicciuca:
All tests passing. Changed the signature of one helper method to simplify the code a bit.

### DIFF
--- a/unordered_associative_task_2/student/src/impl.cpp
+++ b/unordered_associative_task_2/student/src/impl.cpp
@@ -25,11 +25,68 @@ auto xx_hash = [] (const std::string &file)
     return xxh::xxhash<64>(buffer);
 };
 
+/** @todo With help of std filesystem, lists all regular files under specified directory recursively*/
+std::vector<std::string> listFiles(const std::string& directory)
+{
+    std::vector<std::string> fileList;
+
+    for (auto& it : fs::recursive_directory_iterator(directory))
+    {
+        if (it.path().has_extension())
+        {
+            fileList.push_back(it.path().string());
+        }
+    }
+
+    return fileList;
+}
+
+/** @todo Implement function that will remove group in if it has only one or zero elements */
+void removeUniqueGroups(filtering_map& filteredData)
+{
+    for (auto it = filteredData.begin(); it != filteredData.end();)
+    {
+        if (it->second.size() <= 1)
+        {
+            it = filteredData.erase(it);
+        }
+        else
+        {
+            it++;
+        }
+    }
+}
+
+/** @todo Implement function that will transform map to a vector */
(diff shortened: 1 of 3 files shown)


unordered_associative_task_2/student/src/impl.cpp
```diff
@@ -25,11 +25,68 @@ auto xx_hash = [] (const std::string &file)
...
 };
 
+/** @todo With help of std filesystem, lists all regular files under specified directory recursively*/
+std::vector<std::string> listFiles(const std::string& directory)
+{
+    std::vector<std::string> fileList;
+
+    for (auto& it : fs::recursive_directory_iterator(directory))
+    {
+        if (it.path().has_extension())
```
smay1613, line 35: fs::is_regular_file?

unordered_associative_task_2/student/src/impl.cpp
```diff
@@ -25,11 +25,68 @@ auto xx_hash = [] (const std::string &file)
...
+std::vector<std::vector<std::string>> flattenGrouped (const filtering_map& grouped)
+{
+    std::vector<std::vector<std::string>> flatGroups;
+
+    std::vector<std::string> group;
+
+    for (auto& it : grouped)
+    {
+        group.assign(it.second.begin(), it.second.end());
+        flatGroups.push_back(group);
```
smay1613, line 70: why can't you just use .insert() with two iterators, but rather assign to a temp vector?

unordered_associative_task_2/student/src/impl.cpp
```diff
@@ -41,9 +98,21 @@ filtering_map groupDuplicates (const std::vector<std::string>& dataSource, Filte
...
+ * @return list of duplicated files, grouped by content
+ */
 std::vector<std::vector<std::string> > findDuplicates(const std::string &rootPath)
 {
-    // filter by size
-    // filter by content
-    // flatten
+    auto fileList = listFiles(rootPath);
+
+    auto grouped = groupDuplicates(fileList,
```
smay1613, line 110: also please group by size first as an optimization
see how it should look like:

```
    const std::vector<std::string>& files = listFiles(rootPath);
    auto get_size = [](const std::string& file) {
        std::error_code success;
        return fs::file_size(file, success);
    };

    auto sizeGroups = groupDuplicates(files, get_size);
    auto duplicatedSizes = flattenGrouped(sizeGroups);

    auto hashGroups = groupDuplicates(duplicatedSizes, xx_hash);
    std::vector<std::vector<std::string>> duplicates;
    for (const auto& sameHashGroup : hashGroups)
    {
        duplicates.push_back(sameHashGroup.second);
    }

    return duplicates;

```

PR #475, opened 2023-02-14
[Carina Ganea] unordered_associative_task_2

carina-eliza-ganea:
All tests pass.

### DIFF
--- a/unordered_associative_task_2/student/include/impl.h
+++ b/unordered_associative_task_2/student/include/impl.h
@@ -47,10 +47,8 @@ std::vector<std::string> flattenGrouped (const filtering_map& grouped){
 
     std::vector<std::string> flattened;
 
-    for( auto group : grouped){
-        for( auto s : group.second){
-            flattened.push_back(s);
-        }      
+    for( const auto& group : grouped){
+        flattened.insert(flattened.end(), group.second.begin(), group.second.end());      
     }
 
     return flattened;
--- a/unordered_associative_task_2/student/include/impl.h
+++ b/unordered_associative_task_2/student/include/impl.h
@@ -2,17 +2,59 @@
 #include <vector>
 #include <string>
 #include <unordered_map>
+#include <filesystem>
+#include <iostream>
+#include <fstream>
+#include "xxhash.h"
 
 using filtering_map = std::unordered_map<size_t, std::vector<std::string>>;
 
 /** @todo With help of std filesystem, lists all regular files under specified directory recursively*/
-std::vector<std::string> listFiles(const std::string& directory);
+std::vector<std::string> listFiles(const std::string& directory){
+    std::vector<std::string> file_list;
+
+    for (const auto& dir_entry : std::filesystem::recursive_directory_iterator(directory)){
+
+        if(dir_entry.is_regular_file()){
+
+            file_list.push_back(dir_entry.path());
+
+        } else{
+            
+            auto sub_files = listFiles(dir_entry.path());
+        }
+    }
+
+    return file_list;
+}
 
 /** @todo Implement function that will remove group in if it has only one or zero elements */
-void removeUniqueGroups(filtering_map& filteredData);
+void removeUniqueGroups(filtering_map& filteredData){
+
+    for( auto group = filteredData.begin(), last = filteredData.end();  group != last;){
+        
+        if(group->second.size() <= 1){
+
+            group = filteredData.erase(group);
+        } else {
+            group++;
+        }
+    }
+}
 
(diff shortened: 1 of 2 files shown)


unordered_associative_task_2/student/include/impl.h
```diff
@@ -2,17 +2,59 @@
...
+    }
+}
 
 /** @todo Implement function that will transform map to a vector */
-std::vector<std::string> flattenGrouped (const filtering_map& grouped);
+std::vector<std::string> flattenGrouped (const filtering_map& grouped){
+
+    std::vector<std::string> flattened;
+
+    for( auto group : grouped){
```
smay1613: please don't copy the string on each iteration, use const auto& for non-literal types

unordered_associative_task_2/student/include/impl.h
```diff
@@ -2,17 +2,59 @@
...
+        for( auto s : group.second){
```
smay1613: you may avoid this loop if you'll use insert method with two iterators

unordered_associative_task_2/student/src/impl.cpp
```diff
@@ -44,6 +69,19 @@ filtering_map groupDuplicates (const std::vector<std::string>& dataSource, Filte
 std::vector<std::vector<std::string> > findDuplicates(const std::string &rootPath)
 {
     // filter by size
-    // filter by content
+    filtering_map by_size = groupDuplicates(listFiles(rootPath), true);
```
smay1613: FilterCriteria should be a functional object!

unordered_associative_task_2/student/src/impl.cpp
```diff
@@ -30,6 +30,31 @@ auto xx_hash = [] (const std::string &file)
...
+    filtering_map unique;
+
+    for(auto path : files){
+        std::size_t key;
+
+        if(filter){
+
+            std::ifstream stream {path};
+
+            if (!stream)
```
smay1613: why should you open this file?

unordered_associative_task_2/student/src/impl.cpp
```diff
@@ -30,6 +30,31 @@ auto xx_hash = [] (const std::string &file)
...
+            {
+                std::cerr << "File " << path << " can't be opened" << std::endl;
+            }
+
+            key = (!stream)? 0 : fs::file_size(path);
```
smay1613: that must be hidden under the filter

unordered_associative_task_2/student/src/impl.cpp
```diff
@@ -30,6 +30,31 @@ auto xx_hash = [] (const std::string &file)
...
+        } else {
+
+            key = xx_hash(path);
+        }
+
+        std::vector<std::string> none = {};
```
smay1613: the same as a line below

unordered_associative_task_2/student/src/impl.cpp
```diff
@@ -30,6 +30,31 @@ auto xx_hash = [] (const std::string &file)
...
+        unique.try_emplace(key, none);
+        unique[key].push_back(path);
```
smay1613: yep, but please 
`for (const auto& file : files)
    {
        filtered[filter(file)].push_back(file);
    }`

unordered_associative_task_2/student/src/impl.cpp
```diff
@@ -44,6 +69,19 @@ filtering_map groupDuplicates (const std::vector<std::string>& dataSource, Filte
...
+    filtering_map by_size = groupDuplicates(listFiles(rootPath), true);
+
     // flatten
+    std::vector<std::string> duplicated_file_list = flattenGrouped(by_size);
+
+    // filter by content
+    filtering_map by_hash = groupDuplicates(duplicated_file_list, false);
+    
+    std::vector<std::vector<std::string>> duplicated;
+    for( auto group : by_hash){
```
smay1613: by default, auto will be resolved as a plain type name (not a reference!), so add const auto&

unordered_associative_task_2/student/src/impl.cpp
```diff
@@ -25,34 +26,27 @@ auto xx_hash = [] (const std::string &file)
...
+};
 
 /** @todo Implement filtering files by different buckets by specified criteria*/
 template<class FilterCriteria>
 filtering_map filter(const std::vector<std::string>& files, FilterCriteria filter)
 {
     filtering_map unique;
-
+    
     for(auto path : files){
```
smay1613, line 47: const ref here also

PR #476, opened 2023-02-14
[Ionut Demsa] unord_assoc_task_1

Ionut95:


unordered_associative_task_1/student/include/impl.h
```diff
@@ -32,4 +34,23 @@ using playlist = std::vector<Song>;
  *
  * @return true if first playlist is anagram of second
  */
-bool is_same_content(const playlist& first_playlist, const playlist& second_playlist);
+bool is_same_content(const playlist& first_playlist, const playlist& second_playlist) {
+    if (first_playlist.size() != second_playlist.size()) {
+        return false;
+    }
+    std::unordered_multiset<std::string> playlist1 {}, playlist2 {};
```
smay1613: please use Song abstraction instead of std::string

unordered_associative_task_1/student/include/impl.h
```diff
@@ -32,4 +34,23 @@ using playlist = std::vector<Song>;
...
+    for(auto const& song: first_playlist) {
+        playlist1.insert(song.name());
+    }
+    for(auto const& song: second_playlist) {
+        playlist2.insert(song.name());
+    }
+    for(auto const& song: playlist1) {
+        auto it = std::find(playlist2.begin(), playlist2.end(), song);
```
smay1613: please avoid this linear search

Ionut95: I tried to think of another way to do it like to use count to see if all the songs from first playlist are contained in the second one. But also using count method I will need to iterate through one of the playlists. Could you please provide me with an advice in order to avoid the use of this linear search? Thanks!

unordered_associative_task_1/student/include/impl.h
```diff
@@ -32,4 +44,17 @@ using playlist = std::vector<Song>;
  *
  * @return true if first playlist is anagram of second
  */
-bool is_same_content(const playlist& first_playlist, const playlist& second_playlist);
+bool is_same_content(const playlist& first_playlist, const playlist& second_playlist) {
+    if (first_playlist.size() != second_playlist.size()) {
+        return false;
+    }
+    std::unordered_multiset<Song> playlist1;
```
smay1613, line 51: you may use constructor from two iterators here, don't use that range for

unordered_associative_task_1/student/include/impl.h
smay1613: See https://en.cppreference.com/w/cpp/container/unordered_multiset/operator_cmp
It will do it automatically, using the equal_range function

PR #477, opened 2023-02-15
[Carina Ganea] pseudocontainers_task

carina-eliza-ganea:
All tests pass

### DIFF
--- a/pseudocontainers_task/student/include/impl.h
+++ b/pseudocontainers_task/student/include/impl.h
@@ -2,6 +2,8 @@
 #include <string>
 #include <map>
 #include <unordered_map>
+#include <bitset>
+#include <iostream>
 
 using calendar = std::string; // format: o - available, x - vacation
 using worker_id = size_t;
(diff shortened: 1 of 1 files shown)


pseudocontainers_task/student/include/impl.h
```diff
@@ -16,4 +18,28 @@ constexpr char vacationMarker = 'x';
  * @param workerVacations - pairs of worker and his calendar
  * @return Max intersection of two workers and intersection days count
  */
-std::tuple<worker_id, worker_id, days> getMaxIntersection(const std::unordered_map<worker_id, calendar>& workerVacations);
+std::tuple<worker_id, worker_id, days> getMaxIntersection(const std::unordered_map<worker_id, calendar>& workerVacations){
+    days max_intersection = 0;
+    worker_id worker1, worker2;
+
+    for( auto worker: workerVacations){
```
smay1613, line 25: use const ref by default please
you should use auto without reference only when you're dealing with literal types

PR #478, opened 2023-02-15
[Giuseppe Picciuca] pseudocontainers_task

gpicciuca:


pseudocontainers_task/student/include/impl.h
```diff
@@ -10,10 +13,95 @@ constexpr days daysInMonth = 30;
 constexpr char availableMarker = 'o';
 constexpr char vacationMarker = 'x';
 
+size_t getMaxIntersectionCount(const calendar& workerVacation1, const calendar& workerVacation2)
```
smay1613, line 16: Try to use std::bitset here and you'll see how it will simplify the code

PR #480, opened 2023-02-15
[Alberto Icardi] - Pseudocontainer task 

AlbertoIcardi97:


pseudocontainers_task/student/src/impl.cpp
```diff
@@ -1 +1,31 @@
 #include "impl.h"
+#include <bitset>
+std::tuple<worker_id, worker_id, days> getMaxIntersection(const std::unordered_map<worker_id, calendar>& workerVacations){
+
+    worker_id resultWorker1, resultWorker2;
```
smay1613: please default-initialize with {}

pseudocontainers_task/student/src/impl.cpp
```diff
@@ -1 +1,31 @@
...
+    days resultDaysOverlap=0;
+
+
+    for (auto & itOuter : workerVacations){
+
+        std::bitset<daysInMonth> bitsetOuter(itOuter.second,0,itOuter.second.size(),'o','x');
```
smay1613: re-consider naming, you already have a bitset as a type, why should you duplicate it in the name?

PR #481, opened 2023-02-15
[Vincenzo Spinello] pseudocontainers_task

vspinell:


pseudocontainers_task/student/src/impl.cpp
```diff
@@ -1 +1,29 @@
 #include "impl.h"
+#include <bitset>
+
+bool isBiggerIntersection(const calendar& c1,const  calendar& c2, days& maxIntersectionCount) {
+	std::bitset<daysInMonth> bitFormatCalendar1(c1, 0, c1.size(),availableMarker, vacationMarker);
+	std::bitset<daysInMonth> bitFormatCalendar2(c2, 0, c2.size(),availableMarker, vacationMarker);
+	days IntersectionCount =  (bitFormatCalendar1 & bitFormatCalendar2).count();
+	if (IntersectionCount > maxIntersectionCount) {
```
smay1613: std::max?

pseudocontainers_task/student/src/impl.cpp
```diff
@@ -1 +1,29 @@
...
+	return false;
+}
+
+std::tuple<worker_id, worker_id, days> getMaxIntersection(const std::unordered_map<worker_id, calendar>& workerVacations) {
+	days maxIntersectionCount = 0;
+	std::tuple<worker_id, worker_id, days> maxIntersection;
+	if (!workerVacations.empty())
+	{
+		for (auto it = workerVacations.begin();  std::next(it) != workerVacations.end() ; ++it) {
+			for (auto it2 = std::next(it); it2 != workerVacations.end() ; ++it2) {
```
smay1613: looks good, but please detalize why do you need  that next iterator
(do it with a good naming)

pseudocontainers_task/student/src/impl.cpp
```diff
@@ -5,7 +5,7 @@ bool isBiggerIntersection(const calendar& c1,const  calendar& c2, days& maxInter
 	std::bitset<daysInMonth> bitFormatCalendar1(c1, 0, c1.size(),availableMarker, vacationMarker);
 	std::bitset<daysInMonth> bitFormatCalendar2(c2, 0, c2.size(),availableMarker, vacationMarker);
 	days IntersectionCount =  (bitFormatCalendar1 & bitFormatCalendar2).count();
-	if (IntersectionCount > maxIntersectionCount) {
+	if (std::max(IntersectionCount, maxIntersectionCount) != maxIntersectionCount) {
```
smay1613: not this style, try just do:
maxIntersectionCount = std::max(IntersectionCount, maxInstersectionCout)

vspinell: but I also want to return true. Just assigning will cause the condition to be always true

smay1613: so please don't assign and check the condition in one function
consider to change design a little bit

PR #483, opened 2023-02-15
[Andrei Tulpan] unordered_associative_task_2

andreitulpan:


unordered_associative_task_2/student/src/impl.cpp
```diff
@@ -41,9 +74,25 @@ filtering_map groupDuplicates (const std::vector<std::string>& dataSource, Filte
...
+    std::vector<std::string> filesList = flattenGrouped(filesBySize);
+
     // filter by content
+    filtering_map filesByContent = groupDuplicates(filesList,
+        [](const std::string& file) { return xx_hash(file); });
+
     // flatten
+    std::vector<std::vector<std::string>> filesGroupedByContent;
+    for (auto const& fileAndContent : filesByContent) {
+        if (fileAndContent.second.size() > 0) {
```
smay1613, line 92: _!empty()_

unordered_associative_task_2/student/src/impl.cpp
```diff
@@ -41,9 +74,25 @@ filtering_map groupDuplicates (const std::vector<std::string>& dataSource, Filte
...
+            std::vector<std::string> files(fileAndContent.second.begin(), fileAndContent.second.end());
+            filesGroupedByContent.push_back(files);
```
smay1613, line 94: why do you need that temp vector?

PR #487, opened 2023-02-17
[Vincenzo Spinello] iterator_task_1

vspinell:


iterators_task_1/student/src/impl.cpp
```diff
@@ -1,2 +1,14 @@
 #include "impl.h"
 
+std::string reverseWords(const std::string& input) {
+	std::stringstream ss(input);
+	std::string reversed, word;
+	while (ss >> word) {
```
smay1613, line 6: that's acceptable way, but try to achieve this via the algorithms + iterators now

PR #488, opened 2023-02-17
[Giuseppe Picciuca] iterators_task_1

gpicciuca:
Didn't think of std::string::pop_back() before committing so i went with the substr route..

### DIFF
--- a/iterators_task_1/student/include/impl.h
+++ b/iterators_task_1/student/include/impl.h
@@ -1,6 +1,8 @@
 #pragma once
 #include <sstream>
 #include <iterator>
+#include <algorithm>
+#include <vector>
 
 /**
  * @todo Implement function that will reverse words in a sentence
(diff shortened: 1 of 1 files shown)


iterators_task_1/student/include/impl.h
```diff
@@ -10,4 +12,28 @@
...
  * @warning YOU MUST USE STRING STREAMS AND STREAM ITERATORS!
  */
-std::string reverseWords(const std::string& input);
+std::string reverseWords(const std::string& input)
+{
+    std::vector<std::string> token;
+    std::istringstream in(input);
+    std::ostringstream out;
+    
+    std::move(
```
smay1613, line 21: no profit from move here
consider:
stream data: "abc"
you're adding "abd"
how can we move "abd" to "abc"? no way, we must copy

iterators_task_1/student/include/impl.h
```diff
@@ -10,4 +12,28 @@
  *
  * @warning YOU MUST USE STRING STREAMS AND STREAM ITERATORS!
  */
-std::string reverseWords(const std::string& input);
+std::string reverseWords(const std::string& input)
+{
+    std::vector<std::string> token;
+    std::istringstream in(input);
+    std::ostringstream out;
```
smay1613, line 19: please declare variables near the point of usage

iterators_task_1/student/include/impl.h
```diff
@@ -10,4 +12,28 @@
...
+
+    std::copy(
+        token.rbegin(), 
+        token.rend(),
+        std::ostream_iterator<std::string>(out, " ")
+    );
+
+    std::string result = out.str();
+
+    return std::move(result.substr(0, result.length() - 1));
```
smay1613, line 38: you already returning rvalue, why do you need move here?
and yes, pop_back will be better here

PR #491, opened 2023-02-17
[Carina Ganea] iterators_task_1

carina-eliza-ganea:


iterators_task_1/student/include/impl.h
```diff
@@ -10,4 +11,18 @@
...
+
+std::string reverseWords(const std::string& input){
+    std::stringstream ss(input);
+
+    std::string res, word;
+
+    ss >> word;
+    res = word;
+
+    while (ss >> word) {
```
smay1613: acceptable way, but try to solve with help of containers/iterators and algorithms

PR #492, opened 2023-02-17
[Carina Ganea] iterators_task_2

carina-eliza-ganea:


iterators_task_2/student/include/impl.h
```diff
@@ -5,17 +5,42 @@ template <class Container>
 struct accumulator
 {
     /** @todo Iterator traits*/
+    using iterator_category	= typename std::output_iterator_tag;
```
smay1613, line 8: usually for output iterators only tag is provided, others are set to void

iterators_task_2/student/include/impl.h
```diff
@@ -5,17 +5,42 @@ template <class Container>
...
+    using pointer = typename Container::pointer;
+    using reference	= typename Container::reference;
+
     /** @todo Constructor with container*/
+    accumulator(Container& container) : m_container{container} {}
+    
     /** @todo lvalue writing operator*/
+    template<typename T>
+    accumulator<Container>& operator=(const T& value){
+            std::copy(value.begin(), value.end(), std::back_inserter(m_container));
```
smay1613: I'd use insert here

PR #493, opened 2023-02-17
[Giuseppe Picciuca] iterators_task_3

gpicciuca:
The implementation is unfinished as i've come to a blocking point.

If I uncomment the test Usage.BasicAlgorithm, specifically the part with back_inserter into std::list, then the code compiles. 

With it, it doesn't compile due to this error:
`no match for ‘operator=’ (operand types are ‘std::back_insert_iterator<std::__cxx11::list<long unsigned int> >’ and ‘std::pair<const std::__cxx11::basic_string<char>, long unsigned int>’)`

It complains at this line of code in the STL:
```
template<typename _BidirectionalIterator, typename _OutputIterator>
    _OutputIterator
    reverse_copy(_BidirectionalIterator __first, _BidirectionalIterator __last,
		 _OutputIterator __result)
    {
      // concept requirements
      __glibcxx_function_requires(_BidirectionalIteratorConcept<
				  _BidirectionalIterator>)
      __glibcxx_function_requires(_OutputIteratorConcept<_OutputIterator,
		typename iterator_traits<_BidirectionalIterator>::value_type>)
      __glibcxx_requires_valid_range(__first, __last);

      while (__first != __last)
	{
	  --__last;
	  *__result = *__last;     // <==== HERE
	  ++__result;
	}
      return __result;
    }
```
Uncommenting this test, it compiles, but crashes in the 
```
operator value_type()
{
        return m_iter->first;
}
```

implementation in each pair_iterator class because it cannot dereference m_iter (I'm guessing it's the end() iterator, not sure).

Lastly, in maintest.h the std::map has keys of type std::string, but 
`using key_type = size_t;`
so the test TypesAndTraits.ValueTypeTraits fails due to mismatched key_types after implementing the type-traits in pair_iterator. 
Changing the above line to std::string fixes it.

Feedback is well appreciated. Thank you!

### DIFF
--- a/iterators_task_3/student/include/impl.h
+++ b/iterators_task_3/student/include/impl.h
@@ -14,9 +14,49 @@ struct pair_iterator : public Iterator
 template<class Iterator, class T>
 struct pair_iterator<Iterator, true, T> : public Iterator
 {
-    /** @todo Iterator traits */
-    /** @todo Iterator operations */
-    /** @todo Constructor from original iterator */
+    public:
+        /** @todo Iterator traits */
+        using value_type = const std::string;
+        using reference = typename Iterator::reference;
+        using pointer = typename Iterator::pointer;
+        using iterator_category = typename Iterator::iterator_category;
+        using difference_type = typename Iterator::difference_type;
+
+        /** @todo Iterator operations */
+        pair_iterator& operator++()
+        {
+            m_iter++;
+            return *this;
+        }
+
+        pair_iterator& operator--()
+        {
+            m_iter--;
+            return *this;
+        }
+
+        value_type& operator*()
+        {
+            return m_iter->first;
+        }
+
+        template<typename _T>
+        void operator=(const _T& val) = delete;
+        
+        operator value_type()
+        {
+            return m_iter->first;
+        }
+
+        /** @todo Constructor from original iterator */
+        pair_iterator() = default;
+
+        pair_iterator(Iterator it)
+            : m_iter(it)
+        {}
+
+    private:
+        Iterator m_iter;
 };
 
 /**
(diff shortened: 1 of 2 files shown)


iterators_task_3/student/include/impl.h
```diff
@@ -14,9 +14,49 @@ struct pair_iterator : public Iterator
 template<class Iterator, class T>
 struct pair_iterator<Iterator, true, T> : public Iterator
 {
-    /** @todo Iterator traits */
-    /** @todo Iterator operations */
-    /** @todo Constructor from original iterator */
+    public:
+        /** @todo Iterator traits */
+        using value_type = const std::string;
```
smay1613, line 19: stick to the typename T::first_type

iterators_task_3/student/include/impl.h
```diff
@@ -14,9 +14,49 @@ struct pair_iterator : public Iterator
...
+        using reference = typename Iterator::reference;
+        using pointer = typename Iterator::pointer;
+        using iterator_category = typename Iterator::iterator_category;
+        using difference_type = typename Iterator::difference_type;
+
+        /** @todo Iterator operations */
+        pair_iterator& operator++()
```
smay1613, line 26: you already inherit from the iterator, so you shouldn't provide these operations

iterators_task_3/student/include/impl.h
```diff
@@ -14,9 +14,49 @@ struct pair_iterator : public Iterator
...
+        
+        operator value_type()
+        {
+            return m_iter->first;
+        }
+
+        /** @todo Constructor from original iterator */
+        pair_iterator() = default;
+
+        pair_iterator(Iterator it)
```
smay1613, line 54: I agree that it is better to adapt the interface, but here please just change the access type for that iterator

iterators_task_3/student/include/impl.h
```diff
@@ -25,9 +65,59 @@ struct pair_iterator<Iterator, true, T> : public Iterator
 template<class Iterator, class T>
 struct pair_iterator<Iterator, false, T> : public Iterator
 {
-    /** @todo Iterator traits */
-    /** @todo Iterator operations */
-    /** @todo Constructor from original iterator */
+    public:
+        /** @todo Iterator traits */
+        using value_type = size_t;
```
smay1613, line 70: stick please to the T

smay1613:
That compilation error shows you that types are mismatching, so no conversion occur somewhere, so maybe your implementation missing something. 
Iterator expects size_t while your iterator dereferencing returns pair.

smay1613:
Lastly, in maintest.h the std::map has keys of type std::string, but
using key_type = size_t;
so the test TypesAndTraits.ValueTypeTraits fails due to mismatched key_types after implementing the type-traits in pair_iterator.
Changing the above line to std::string fixes it.

I may be wrong, but actually that is a test about not hardcoding the traits

PR #494, opened 2023-02-17
[Alberto Icardi] - Iterators task 1 

AlbertoIcardi97:


iterators_task_1/student/include/impl.h
```diff
@@ -10,4 +12,21 @@
  *
  * @warning YOU MUST USE STRING STREAMS AND STREAM ITERATORS!
  */
-std::string reverseWords(const std::string& input);
+std::string reverseWords(const std::string& input){
+    std::stringstream in{input};
+    std::stringstream out{};
```
smay1613: please always declare variables near the point of usage, don't declare everything in one place (see cpp core guidelines)

iterators_task_1/student/include/impl.h
```diff
@@ -10,4 +12,21 @@
...
+    std::vector<std::string> token{std::istream_iterator<std::string> {in},std::istream_iterator<std::string> {}};
+    
+    if(token.empty()){
+        return "";
+    }
+    std::move(token.rbegin(),token.rend(),std::ostream_iterator<std::string> {out, " "});
```
smay1613: no profit from move here, data is accumulated in the stream

PR #497, opened 2023-02-19
[Andrei Tulpan] iterators_task_2

andreitulpan:


iterators_task_2/student/include/impl.h
```diff
@@ -4,12 +4,38 @@
 template <class Container>
 struct accumulator
 {
+protected:
+    Container& m_container;
+
+public:
     /** @todo Iterator traits*/
+    typedef typename Container::value_type value_type;
```
smay1613, line 12: 1. Prefer using
2. Usually traits are void (only tag is provided) for output iterators

iterators_task_2/student/include/impl.h
```diff
@@ -4,12 +4,38 @@
...
+    typedef typename Container::pointer pointer;
+    typedef typename Container::reference reference;
+
     /** @todo Constructor with container*/
+    accumulator(Container& otherContainer) : m_container(otherContainer) {}
+
     /** @todo lvalue writing operator*/
+    template <class OtherContainer>
+    accumulator& operator=(OtherContainer const& input) {
+        std::copy(input.cbegin(), input.cend(), std::back_inserter(m_container));
```
smay1613, line 24: so maybe it's better to use insert here?

PR #498, opened 2023-02-20
[Alberto Icardi]- Algorithms task 1

AlbertoIcardi97:


algorithms_task_1/student/include/impl.h
```diff
@@ -11,4 +11,12 @@
   * @return consecutive range
 */
 template<class FI, class Comparator = std::equal_to<typename std::iterator_traits<FI>::value_type>>
-std::pair<FI, FI> consecutive_group(FI start, FI end, Comparator comp = {});
+std::pair<FI, FI> consecutive_group(FI start, FI end, Comparator comp = {}){
+  auto first=std::adjacent_find(start,end, comp);
+  auto second=std::adjacent_find(first,end, std::not_equal_to<typename std::iterator_traits<FI>::value_type>());
```
smay1613: and what if comparator is not an std::equal?

algorithms_task_1/student/include/impl.h
```diff
@@ -13,7 +13,12 @@
 template<class FI, class Comparator = std::equal_to<typename std::iterator_traits<FI>::value_type>>
 std::pair<FI, FI> consecutive_group(FI start, FI end, Comparator comp = {}){
   auto first=std::adjacent_find(start,end, comp);
-  auto second=std::adjacent_find(first,end, std::not_equal_to<typename std::iterator_traits<FI>::value_type>());
+
+  auto not_comp= [&](const typename std::iterator_traits<FI>::value_type & a,const typename std::iterator_traits<FI>::value_type &  b)
```
smay1613, line 17: also you may try std::not2 and std::not_fn

algorithms_task_1/student/include/impl.h
```diff
@@ -13,7 +13,12 @@
...
+  {return !comp(a,b);};
+  // Since C++17 I could have used std::not_fn(comp) !!!
```
smay1613, line 19: you may try std::not2, but that's not good idea

AlbertoIcardi97: So, is my way correct?

smay1613: Yes!

PR #499, opened 2023-02-20
[Vincenzo Spinello ] algorithm_task_1

vspinell:


vspinell:
sorry for the param name of the lambda...I was testing and I forgot to change its name. I would have rather called it "val".

algorithms_task_1/student/include/impl.h
```diff
@@ -11,4 +11,11 @@
   * @return consecutive range
 */
 template<class FI, class Comparator = std::equal_to<typename std::iterator_traits<FI>::value_type>>
-std::pair<FI, FI> consecutive_group(FI start, FI end, Comparator comp = {});
+std::pair<FI, FI> consecutive_group(FI start, FI end, Comparator comp = {}) {
+	auto itBegin = std::adjacent_find(start, end, comp);
+	if (itBegin != end) {
+		auto itEnd = std::find_if_not(itBegin, end, [itBegin](typename std::iterator_traits<FI>::value_type bo){return bo == *itBegin;});
```
smay1613: please dont copy the value, mark it as a const reference

PR #501, opened 2023-02-20
[Giuseppe Picciuca] algorithms_task_1

gpicciuca:


algorithms_task_1/student/include/impl.h
```diff
@@ -10,5 +11,22 @@
...
+{
+  auto first = std::adjacent_find(start, end, comp);
+  
+  if (first == end)
+    return {end, end};
+
+  auto val = *first;
+
+  auto last = std::find_if_not(first, end, 
+    [&](T element)
```
smay1613, line 25: please don't copy the value to lambda

algorithms_task_1/student/include/impl.h
```diff
@@ -10,5 +11,22 @@
   * @param comparator - function to compare for equality
   * @return consecutive range
 */
-template<class FI, class Comparator = std::equal_to<typename std::iterator_traits<FI>::value_type>>
-std::pair<FI, FI> consecutive_group(FI start, FI end, Comparator comp = {});
+template<class FI, class T = typename std::iterator_traits<FI>::value_type, class Comparator = std::equal_to<T>>
+std::pair<FI, FI> consecutive_group(FI start, FI end, Comparator comp = {})
+{
+  auto first = std::adjacent_find(start, end, comp);
```
smay1613, line 17: so can't we use plain find here? why do we need adjacent_find?

gpicciuca: Because find() searches for the first position of a given element. The first position doesn't necessarily mean that the next element will be the same as the previous one so adjacent_find checks that for us and returns the iterator to the element whose next element is the same. With std::find_if_not we find the next mismatch of the sequence so we get a first and last iterator for the sequence we need.

PR #502, opened 2023-02-20
[Giuseppe Picciuca] algorithms_task_2

gpicciuca:


algorithms_task_2/student/include/impl.h
```diff
@@ -9,4 +9,7 @@
  *
  * @return true if source starts from a prefix
  */
-bool startsWith(const std::string& source, const std::string& prefix);
+bool startsWith(const std::string& source, const std::string& prefix)
+{
+     return source.find(prefix) == 0;
```
smay1613: will be not effective if prefix is not present

gpicciuca:
Please ignore the wrong commit. Too many branches.. 

Either way, I rewrote the function and added 3 possible ways in my opinion how this method could be implemented. The last 2 actually return the same result since string::find returns the position inside the string where the searched pattern begins so if it returns 0 we're actually at the beginning of the string. An empty prefix matches at the beginning of the string, too..

algorithms_task_2/student/include/impl.h
```diff
@@ -9,4 +9,9 @@
  *
  * @return true if source starts from a prefix
  */
-bool startsWith(const std::string& source, const std::string& prefix);
+bool startsWith(const std::string& source, const std::string& prefix)
+{
+     return source.rfind(prefix, source.find_first_not_of(' ')) == 0; // ignore initial whitespaces
```
smay1613, line 14: I'd use equal here for bool, or mismatch for iterators.
Find will not stop if case if char mismatch

PR #503, opened 2023-02-20
[Bogdan Pana] iterators_task_1

BogdanPana26:


iterators_task_1/student/include/impl.h
```diff
@@ -10,4 +11,19 @@
  *
  * @warning YOU MUST USE STRING STREAMS AND STREAM ITERATORS!
  */
-std::string reverseWords(const std::string& input);
+std::string reverseWords(const std::string& input)
+{
+    std::stringstream Stream(input);
+    std::stringstream Out;
+    std::istream_iterator<std::string> begin(Stream);
```
smay1613, line 18: inline usage is better as for me, create it in during constructor call

PR #505, opened 2023-02-20
[Bogdan Pana] algorithms_task_1

BogdanPana26:


algorithms_task_1/student/include/impl.h
```diff
@@ -11,4 +11,13 @@
...
 template<class FI, class Comparator = std::equal_to<typename std::iterator_traits<FI>::value_type>>
-std::pair<FI, FI> consecutive_group(FI start, FI end, Comparator comp = {});
+std::pair<FI, FI> consecutive_group(FI start, FI end, Comparator comp = {})
+{
+    auto first = std::adjacent_find(start,end,comp);
+    if(first == end)
+    {
+        return std::make_pair(end,end);
+    }
+    auto second = std::find_end(first,end,first,std::next(first),comp);
```
smay1613, line 21: not sure that we should use find_end here
maybe negate a comp?

BogdanPana26: yes I thought of that first , but I don't know how to , the ! operator doesn't work

smay1613: you can use smth like https://en.cppreference.com/w/cpp/utility/functional/not_fn or https://en.cppreference.com/w/cpp/utility/functional/not2

PR #506, opened 2023-02-20
[Bogdan Pana] algorithms_task_2

BogdanPana26:


algorithms_task_2/student/include/impl.h
```diff
@@ -9,4 +9,7 @@
  *
  * @return true if source starts from a prefix
  */
-bool startsWith(const std::string& source, const std::string& prefix);
+bool startsWith(const std::string& source, const std::string& prefix)
+{
+    return source.find(prefix) == 0;
```
smay1613, line 14: will be not effective if prefix is not present

PR #507, opened 2023-02-20
[Vincenzo Spinello] algorithm_task_2

vspinell:


algorithms_task_1/student/include/impl.h
```diff
@@ -11,4 +11,11 @@
   * @return consecutive range
 */
 template<class FI, class Comparator = std::equal_to<typename std::iterator_traits<FI>::value_type>>
-std::pair<FI, FI> consecutive_group(FI start, FI end, Comparator comp = {});
+std::pair<FI, FI> consecutive_group(FI start, FI end, Comparator comp = {}) {
```
smay1613, line 14: please exclude

algorithms_task_2/student/src/impl.cpp
```diff
@@ -1 +1,6 @@
 #include "impl.h"
+#include <algorithm>
+
+bool startsWith(const std::string& source, const std::string& prefix) {
+	return std::search(source.begin(), source.end(), prefix.begin(), prefix.end()) == source.begin();
```
smay1613: will be not effective in case if prefix is not present

algorithms_task_2/student/src/impl.cpp
```diff
@@ -2,5 +2,5 @@
 #include <algorithm>
 
 bool startsWith(const std::string& source, const std::string& prefix) {
-	return std::search(source.begin(), source.end(), prefix.begin(), prefix.end()) == source.begin();
+	return std::equal(source.begin(), source.begin() + prefix.size(), prefix.begin());
```
smay1613, line 5: and what if prefix bigger than a source?

vspinell: It should does not affect the result. The task says "Check if strings starts with another string". So if prefix is a substring of source starting from the first position, it should be fine in my opinion. There is also a test in which prefix is bigger and it passes.

smay1613: source.begin() + size is invalid for example when source is empty

PR #509, opened 2023-02-20
[Carina Ganea] algorithms_task_2

carina-eliza-ganea:


algorithms_task_2/student/include/impl.h
```diff
@@ -9,4 +10,10 @@
  *
  * @return true if source starts from a prefix
  */
-bool startsWith(const std::string& source, const std::string& prefix);
+bool startsWith(const std::string& source, const std::string& prefix){
+    return source.find(prefix, 0) == 0;
```
smay1613: what if prefix will be at the end? so that will be ineffective

algorithms_task_2/student/include/impl.h
```diff
@@ -9,4 +10,10 @@
...
+    // Can also check source.begin() == std::search(source.begin(), source.end(), prefix.begin(), prefix.end());
+    // Another possibility is std::mismatch
+    // Could also check if std::equal(source.begin(), std::next(source.begin(), prefix.size), prefix.begin(), prefix.end()) is true
```
smay1613: so why have you chosen other alternative?

algorithms_task_2/student/include/impl.h
carina-eliza-ganea: The task description says it should start with prefix so I thought I should only check from first position. I didn't realize it could also be a suffix and that I should search the end as well. Could you maybe amend the task description?

smay1613: It should, but it can also be not - that's a software development
I mean that if prefix is not present, this search will be linear, while we can return immediately after comparison of the first chars

algorithms_task_2/student/include/impl.h
carina-eliza-ganea: My main reason was that it would be more readable than the other ways. Complexity-wise I figured they should be almost the same.

algorithms_task_2/student/include/impl.h
carina-eliza-ganea: Oh, I see. Thank you! I'll add a character count as well.

algorithms_task_2/student/include/impl.h
smay1613: std::equal will return on first mismatch, so it's the most effective way

algorithms_task_2/student/include/impl.h
```diff
@@ -11,9 +11,9 @@
  * @return true if source starts from a prefix
  */
 bool startsWith(const std::string& source, const std::string& prefix){
-    return source.find(prefix, 0) == 0;
+    return std::mismatch(source.begin(), std::next(source.begin(), prefix.size()), prefix.begin()) == make_pair(std::next(source.begin(), prefix.size()), prefix.end());
```
smay1613: equal will be simpler for that task I think

smay1613: range size check before?

PR #511, opened 2023-02-20
adaptors_task_ApostolRazvan

ApostolRazvan:


adaptors_task/student/include/impl.h
```diff
@@ -11,4 +13,18 @@
  *
  * @return true if expression is valid
  */
-bool isValid(const std::string& source);
+bool isValid(const std::string& source){
+    std::unordered_map<char, char> brackets = {{')', '('}, {']', '['}, {'}', '{'}};
```
smay1613, line 17: please use container defined in impl.cpp

PR #512, opened 2023-02-20
ordered_associative_task_ApostolRazvan

ApostolRazvan:


ordered_associative_task/student/include/impl.h
```diff
@@ -15,4 +16,20 @@
...
+size_t convertMarkTo5PointSystem(size_t markIn100Point){
+    std::map<size_t, size_t> gradeMap = {
+                {0, 1},
+                {30, 2},
+                {50, 3},
+                {70, 4},
+                {90, 5}
+            };
+
+    for (auto it = gradeMap.rbegin(); it != gradeMap.rend(); ++it) {
```
smay1613, line 28: so there is absolutely no profit from map as you traverse it one by one (linear complexity)
try to use upper_bound here

PR #516, opened 2023-02-20
[Alberto Icardi] - Algorithms task 2

AlbertoIcardi97:


algorithms_task_2/student/include/impl.h
```diff
@@ -9,4 +10,16 @@
  *
  * @return true if source starts from a prefix
  */
-bool startsWith(const std::string& source, const std::string& prefix);
+bool startsWith(const std::string& source, const std::string& prefix){
+    return std::equal(prefix.begin(),prefix.end(),source.begin());
```
smay1613, line 17: please check the size before std::equal call

algorithms_task_2/student/include/impl.h
```diff
@@ -9,4 +10,16 @@
  *
  * @return true if source starts from a prefix
  */
-bool startsWith(const std::string& source, const std::string& prefix);
+bool startsWith(const std::string& source, const std::string& prefix){
+    return std::equal(prefix.begin(),prefix.end(),source.begin()); 
+    // Complexity: prefix.begin()-prefix.end() applications of the predicate
+
+    // Alternatives:
+    // 1) return source.find(prefix)==0;
```
smay1613, line 21: What if prefix will be at the end?

algorithms_task_2/student/include/impl.h
```diff
@@ -9,4 +10,16 @@
...
+    // Complexity: prefix.begin()-prefix.end() applications of the predicate
+
+    // Alternatives:
+    // 1) return source.find(prefix)==0; 
+    //    Complexity: Unspecified, but generally up to linear in length()-pos times the length of the sequence to match (worst case)
+
+    // 2) return std::search(source.begin(),source.end(),prefix.begin(),prefix.end())==source.begin(); 
+    //    Complexity: At most S*N comparisons where S = std::distance(prefix.begin(),prefix.end()) and N = std::distance(source.begin(),source.end())
+
+    // Conclusion: std::equal is the best solution for complexity and for code readability
```
smay1613, line 27: and what is the difference with mismatch?

algorithms_task_2/student/include/impl.h
```diff
@@ -9,4 +10,16 @@
...
-bool startsWith(const std::string& source, const std::string& prefix);
+bool startsWith(const std::string& source, const std::string& prefix){
+    return std::equal(prefix.begin(),prefix.end(),source.begin()); 
+    // Complexity: prefix.begin()-prefix.end() applications of the predicate
+
+    // Alternatives:
+    // 1) return source.find(prefix)==0; 
+    //    Complexity: Unspecified, but generally up to linear in length()-pos times the length of the sequence to match (worst case)
+
+    // 2) return std::search(source.begin(),source.end(),prefix.begin(),prefix.end())==source.begin();
```
smay1613, line 24: will be not effective if prefix is not present

algorithms_task_2/student/include/impl.h
AlbertoIcardi97: Yes, I know if prefix is at the end or also if prefix is not found, the code will take linear time, so it will be ineffective. I think it would be possible to check the first character before calling the find method in order to speed up the code. Anyway this alternative is less efficient and more complicated than the std::equal

algorithms_task_2/student/include/impl.h
AlbertoIcardi97: Maybe that in C++11 mismatch is less safer than search because it does not provide a check on the sizes and assume that the second range has at least many elements as the first

smay1613: same as std::equal
but mismatch will you give iterators, while equal - not

PR #518, opened 2023-02-20
[Iulia Corici] algorithms_task_1

icorici:


algorithms_task_1/student/include/impl.h
```diff
@@ -11,4 +11,11 @@
   * @return consecutive range
 */
 template<class FI, class Comparator = std::equal_to<typename std::iterator_traits<FI>::value_type>>
-std::pair<FI, FI> consecutive_group(FI start, FI end, Comparator comp = {});
+std::pair<FI, FI> consecutive_group(FI start, FI end, Comparator comp = {}) {
+    auto first = std::adjacent_find(start, end, comp);
+    if (first == end) {
+        return std::make_pair(first, first);
+    }
+    auto second = std::find_end(first, end, first, std::next(first), comp);
```
smay1613, line 19: not sure we should use find_end here
consider to perform search on negated comparator

PR #519, opened 2023-02-20
[Iulia Corici] algorithms_task_2

icorici:


algorithms_task_2/student/include/impl.h
```diff
@@ -9,4 +10,6 @@
  *
  * @return true if source starts from a prefix
  */
-bool startsWith(const std::string& source, const std::string& prefix);
+bool startsWith(const std::string& source, const std::string& prefix) {
+    return std::equal(prefix.begin(), prefix.end(), source.begin());
```
smay1613, line 14: you must check sizes before call to std::equal

PR #521, opened 2023-02-20
[Ionut Demsa] algortihms_task_2

Ionut95:


algorithms_task_2/student/include/impl.h
```diff
@@ -9,4 +10,7 @@
  *
  * @return true if source starts from a prefix
  */
-bool startsWith(const std::string& source, const std::string& prefix);
+bool startsWith(const std::string& source, const std::string& prefix) {
+    std::size_t isPrefixOf = source.find(prefix);
```
smay1613: and what if prefix will be present at the end?

algorithms_task_2/student/include/impl.h
```diff
@@ -12,5 +12,5 @@ using std::size_t;
  */
 bool startsWith(const std::string& source, const std::string& prefix) {
     std::size_t isPrefixOf = source.find(prefix);
-    return isPrefixOf != std::string::npos;
+    return isPrefixOf != std::string::npos && isPrefixOf == 0;
```
smay1613: will be not effective if prefix is not present

PR #524, opened 2023-02-20
[Vincenzo Spinello] algorithms_task_3

vspinell:
I tried with the same std::equal test and it succeeded, but one of the tests failed...As far as I understand this is how it should behave and I have no clue how to do it differently.

### DIFF
--- a/algorithms_task_3/student/include/impl.h
+++ b/algorithms_task_3/student/include/impl.h
@@ -17,6 +17,6 @@
 
 template<class BiIt>
 BiIt shift_right_n(BiIt begin, BiIt end, typename BiIt::difference_type shift){
-	std::move_backward(begin, std::prev(end, shift) ,end);
+	std::move_backward(begin, end, std::next(end, shift));
 	return std::next(begin, shift);
 }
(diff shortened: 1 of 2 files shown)


algorithms_task_3/student/include/impl.h
```diff
@@ -12,5 +14,9 @@
  *
  * @note behavior is undefined if shift out of bounds
  */
+
 template<class BiIt>
-BiIt shift_right_n(BiIt begin, BiIt end, typename BiIt::difference_type shift);
+BiIt shift_right_n(BiIt begin, BiIt end, typename BiIt::difference_type shift){
+	std::move_backward(begin, std::prev(end, shift) ,end);
+	return std::next(begin, shift);
```
smay1613, line 21: you should shift the end, while the input range is from begin and end

algorithms_task_3/student/include/impl.h
```diff
@@ -17,6 +17,6 @@
 
 template<class BiIt>
 BiIt shift_right_n(BiIt begin, BiIt end, typename BiIt::difference_type shift){
-	std::move_backward(begin, std::prev(end, shift) ,end);
+	std::move_backward(begin, end, std::next(end, shift));
 	return std::next(begin, shift);
```
smay1613, line 21: may return result of move_backward

PR #531, opened 2023-02-21
[Carina Ganea] algorithms_task_6

carina-eliza-ganea:


algorithms_task_6/student/include/impl.h
```diff
@@ -18,4 +18,7 @@ void insert_to_sorted(Sequence& target, const FwIt sourceBegin, const FwIt sourc
     static_assert(std::is_const<typename std::remove_reference<
                           typename std::iterator_traits<FwIt>::reference>::type>::value,
                       "FwIt must point to const data");
+    for(auto it = sourceBegin; it != sourceEnd; ++it){
+        target.insert(std::upper_bound(target.begin(), target.end(), *it), *it);
```
smay1613, line 22: What if target will be filled with the duplicate values? Will this solution be efficient?

PR #533, opened 2023-02-21
[Andrei Tulpan] algorithms_task_2

andreitulpan:


algorithms_task_2/student/include/impl.h
```diff
@@ -9,4 +9,7 @@
  *
  * @return true if source starts from a prefix
  */
-bool startsWith(const std::string& source, const std::string& prefix);
+bool startsWith(const std::string& source, const std::string& prefix) {
+    auto mismatchPoint = std::mismatch(source.begin(), source.end(), prefix.begin());
```
smay1613, line 17: size check before mismatch?

andreitulpan: done

PR #544, opened 2023-02-21
[Ionut Demsa] algorithms_task_5

Ionut95:


algorithms_task_5/student/include/impl.h
```diff
@@ -10,4 +10,10 @@
  * @return gathered range
  */
 template <typename BiIt, typename UnaryPredicate>
-std::pair<BiIt, BiIt> gather(BiIt begin, BiIt end, BiIt position, UnaryPredicate predicate);
+std::pair<BiIt, BiIt> gather(BiIt begin, BiIt end, BiIt position, UnaryPredicate predicate) {
+    auto negatePredicate = [&predicate](const typename BiIt::value_type& value) {return !predicate(value);};
+    return {
+        std::partition(begin, position, negatePredicate),
```
smay1613, line 16: stable_partition?

Ionut95: I wanted to ask for an alternative solution, does partition_copy would help me with this problem? I know it returns 2 iterators.

smay1613: solution is ok, but here it's better to preserve relative order

PR #548, opened 2023-02-22
[Ionut Demsa] algorithms_task_6

Ionut95:
Hi! I made this solution. But then I thought of another away, much better I think, to solve the task, the idea from the commented line. The thing is I am blocked here with this comparison. Do you think it's possible to overcome this or I should change my idea from the commented line? Thanks.

### DIFF
--- a/algorithms_task_6/student/include/impl.h
+++ b/algorithms_task_6/student/include/impl.h
@@ -18,4 +18,9 @@ void insert_to_sorted(Sequence& target, const FwIt sourceBegin, const FwIt sourc
     static_assert(std::is_const<typename std::remove_reference<
                           typename std::iterator_traits<FwIt>::reference>::type>::value,
                       "FwIt must point to const data");
+    for(auto iter = sourceBegin; iter != sourceEnd; ++iter) {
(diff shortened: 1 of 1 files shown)


algorithms_task_6/student/include/impl.h
```diff
@@ -18,4 +18,9 @@ void insert_to_sorted(Sequence& target, const FwIt sourceBegin, const FwIt sourc
     static_assert(std::is_const<typename std::remove_reference<
                           typename std::iterator_traits<FwIt>::reference>::type>::value,
                       "FwIt must point to const data");
+    for(auto iter = sourceBegin; iter != sourceEnd; ++iter) {
+        auto it = std::upper_bound(target.begin(), target.end(), *iter);
+        target.insert(it, *iter);
+        //auto found = std::find_if(target.begin, target.end(), [&sourceBegin](Sequence& item){return item > static_cast<Sequence>(*sourceBegin);});
```
smay1613, line 24: find is a linear search, not effective
lower bound vs upper bound will make difference if out range contains equal values

